PR #2, opened 2020-08-03
Add base terraform config for batch

jashapiro:
This draft PR adds terraform files to build a the full stack of resources required for a nextflow queue.

This should include the necessary groups, roles, policies, compute environments and batch queues.

Of note, the policies include read/write access to only two buckets: nextflow-ccdl-data and nextflow-ccdl-results (read access is to all buckets).

The compute environments use an AMI based on Amazon ECS-Optimized Amazon Linux AMI, with changes to allow for greater data storage and installation of aws-cli as required for Nextflow.

Most of the settings I used were derived from instructions found here: https://t-neumann.github.io/pipelines/AWS-pipeline/, as well as by looking at the setup created by the cloudformation stack (https://aws.amazon.com/quickstart/biotech-blueprint/nextflow/) that we had previously done some testing with.

I have not yet attempted to deploy this stack, as I wanted to get some feedback/checks before I did so, especially as it is the first time I have used terraform. 😰

While I think most of the stack is probably mostly correct, I am really flying blind on the vpc settings in nextflow-security.tf. I would love to have those looked at especially closely!

I also have not tagged anything, so if there are particular tagging conventions that I should follow, that would be great to hear about. (If there is a way to apply a set of tags to all resources at once, that would probably make the most sense...)


aws/nextflow-compute.tf
```diff
@@ -0,0 +1,56 @@
...
+  compute_environment_name = "nextflow-ondemand-compute"
+  compute_resources {
+    instance_role = aws_iam_instance_profile.nf_ecs_instance_role.arn
+    instance_type = [
+      "optimal",
+    ]
+    allocation_strategy = "BEST_FIT"
+    max_vcpus = 20
+    min_vcpus = 0
+    image_id = "ami-0a8857ac38c35157f"
```
kurtwheeler, line 16: I know you noted where this AMI came from and what it is in the PR description, but I think it'd be good to also note that in a comment here.

aws/nextflow-compute.tf
```diff
@@ -0,0 +1,56 @@
+resource "aws_iam_instance_profile" "nf_ecs_instance_role" {
+  name = "nextflow-ecs-instance-role"
+  role = aws_iam_role.nf_ecs_role.name
+}
+
+resource "aws_batch_compute_environment" "nextflow_ondemand" {
```
kurtwheeler, line 6: It might be nice to standardize on `nextflow_` or `nf_` so you don't have to remember when to use which.

aws/nextflow-compute.tf
kurtwheeler: Oh I just saw `aws/setup-log.md`, maybe just mention that doc in a comment here?

aws/setup-log.md
```diff
@@ -0,0 +1,35 @@
...
+Files for input to workflows (fastq, etc.) will be stored in the `s3://ccdl-scpca-data/raw` path.
+
+## AMI creation
+
+The AMI for batch jobs was set up manually, following the instructions at
+https://www.nextflow.io/docs/latest/awscloud.html#custom-ami
+
+Briefly:
+1. We used the base AMI _Amazon ECS-Optimized Amazon Linux AMI_. (note, not Amazon Linux 2).
+This was launched with an 80GB EBS volume for data (compared to teh default 22) in addition to the base 8GB boot volume.
```
kurtwheeler, line 21:
```suggestion
This was launched with an 80GB EBS volume for data (compared to the default 22) in addition to the base 8GB boot volume.
```

aws/nextflow-compute.tf
```diff
@@ -0,0 +1,56 @@
...
+  compute_resources {
+    instance_role = aws_iam_instance_profile.nf_ecs_instance_role.arn
+    instance_type = [
+      "optimal",
+    ]
+    allocation_strategy = "SPOT_CAPACITY_OPTIMIZED"
+    spot_iam_fleet_role = aws_iam_role.nf_ecs_role.arn
+    bid_percentage = 20
+    max_vcpus = 100
+    min_vcpus = 0
```
kurtwheeler, line 42: Ok so if I understand this correctly, you've got a default queue that uses spot instances and a priority queue that uses on demand instances. They're both going to scale up automatically, default to a max_vcpus of 100 and the priority to 20. Is that all correct?

aws/nextflow-compute.tf
```diff
@@ -0,0 +1,56 @@
...
+    subnets = [
+      aws_subnet.nf_subnet.id,
+    ]
+    type = "EC2"
+  }
+
+  service_role = aws_iam_role.nf_batch_role.arn
+  type         = "MANAGED"
+  depends_on   = [aws_iam_role_policy_attachment.nf_batch]
+}
```
kurtwheeler, line 56: Looks like you're missing newlines at the ends of all your .tf files.

aws/nextflow-policies.tf
```diff
@@ -0,0 +1,173 @@
...
+    "s3:GetBucketAcl",
+    "s3:PutInventoryConfiguration",
+    "s3:GetObjectTorrent",
+    "s3:PutBucketWebsite",
+    "s3:PutBucketRequestPayment",
+    "s3:PutObjectRetention",
+    "s3:GetBucketCORS",
+    "s3:GetBucketLocation",
+    "s3:ReplicateDelete",
+    "s3:GetObjectVersion"
```
kurtwheeler, line 74: Is this everything? I think `s3:*` might be equivalent.

aws/nextflow-security.tf
```diff
@@ -0,0 +1,19 @@
...
+  }
+}
+
+resource "aws_vpc" "nf_vpc" {
+  cidr_block = "10.1.0.0/16"
+}
+
+resource "aws_subnet" "nf_subnet" {
+  vpc_id = aws_vpc.nf_vpc.id
+  cidr_block = "10.1.1.0/24"
```
kurtwheeler, line 18: Are you planning on adding additional subnets? If not, you may just want to have this match the VPC's CIDR block.

aws/nextflow-compute.tf
jashapiro: Yeah, I thought about this before, but with this extra prompting, I will do it.

aws/nextflow-compute.tf
jashapiro: Added.

aws/nextflow-compute.tf
jashapiro: That is what is supposed to happen, yes. I may well change these in the future, but this seemed a good place to start.

aws/nextflow-compute.tf
jashapiro: Added.

aws/nextflow-policies.tf
jashapiro: I don't know? Probably is nearly equivalent, but this was one that I copied from the CloudFormation setting, so I didn't want to mess around too much. I doubt changing it to s3:* would be a problem.

aws/nextflow-security.tf
jashapiro: I don't expect so? This is where I was completely lost, and really just was going off examples. I don't know what the function of multiple subnets would be, but I needed to make at least one, it seemed. the CloudFormation-created vpc definitely made more than one, but I don't know why, or how that might be used.

jashapiro:
> Ok this all looks pretty good to me. You'll definitely want to test it, it's hard to eyeball all the things that could be wrong but there's nothing major as far as I can tell.

That is good to hear. 

What is your recommended test procedure? I have been making sure syntax, etc works with `terraform plan` but I wasn't really sure the best next steps. I assume `terraform apply`, but if things go wrong, what is the rollback? Will `terraform destroy` only affect the resources defined in these files? If I make changes, do I use `apply` again to update things as needed, or is there another step?

aws/nextflow-compute.tf
kurtwheeler: Ok sweet. A summary like that might also be useful to have in a comment or doc somewhere.

aws/nextflow-policies.tf
kurtwheeler: Word if you got this from there then it's probably best to leave it as is.

aws/nextflow-security.tf
kurtwheeler: Ok then this is should be fine. It shouldn't really matter much anyway.

kurtwheeler:
> Will `terraform destroy` only affect the resources defined in these files?

Yep, it's pretty good about that.

> If I make changes, do I use apply again to update things as needed, or is there another step?

There is no other step other than maybe `terraform plan`. `terraform apply` will figure out what it needs to change and only modify/recreate those resources. terraform won't really do anything dangerous. The worst it can really do is get a corrupted state which will make it really hard to cleanup, but that's not that scary and you probably won't run into it.

aws/nextflow-compute.tf
jashapiro: Yeah I was planning to add a readme before making it into a real PR.

kurtwheeler:
Ok I added some untested terraform code. It shouldn't work just yet, you'll need to add the public key of your ssh key into the keypair resource definition. Also https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/batch_compute_environment#ec2_key_pair didn't specify what attribute of the key_pair it wanted, so maybe you'll need to switch that to id (normal instances use the name though).

Once this is set up, you'll be able to ssh in with `ssh -i <private key file> <user name>@<instance ip>`. If you get the user name wrong, it won't be obvious that's the issue. It'll depend on your AMI. I've had to use `ec2_user`, `ubuntu`, and `root` in the past.

If any of that gives you any trouble lemme know!

jashapiro:
After Kurt successfully got things working by adding in missing gateway settings (who knew? Kurt), I have gone back and tightened things back up. I left in the sections that would allow SSH access to nodes, but that access is currently disabled: there is no public key passed to the nodes, and port 22 is closed. 

I have also added more comments in the `aws/setup-log.md` (which maybe should be renamed `README.md`?) to describe the setup in a bit more detail. 

I have also added an example `nextflow.config` file that includes minimal settings for running a job on the default batch queue.

aws/setup-log.md
```diff
@@ -0,0 +1,63 @@
...
+
+The setup of the AWS batch queue and related resources for Nextflow usage are handled by the Terraform (`.tf`) files in this directory.
+
+
+The general setup currently consists of two Batch queues, `nextflow-batch-default-queue`, which includes up to 100 vCPUs and `nextflow-batch-priority-queue`, which includes up to 20 vCPUs.
+`nextflow-batch-default-queue` should be used in all but exceptional circumstances, as it uses spot instances for its processing, at a significant cost savings.
+The spot price threshold is set high (currently 100%!) so it should never not run.
+
+To use these files, you will need to install Terraform on your local machine, either from https://www.terraform.io/downloads.html or via Conda, Homebrew, or the like.
+(You will also probably want the `aws` cli tools, and to have run `aws configure` to set your access key ID and  secret access key.)
```
jaclyn-taroni, line 56:
```suggestion
(You will also probably want the `aws` cli tools, and to have run [`aws configure`](https://docs.aws.amazon.com/cli/latest/reference/configure/index.html) to set your access key ID and  secret access key.)
```

Alternatively link to: https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-quickstart.html

aws/setup-log.md
```diff
@@ -0,0 +1,63 @@
...
+
+Once you have installed terraform, you can run `terraform init` from this directory to get started.
+To implement any changes, you will want a copy of the current state file, which is not included in this repository for security reasons.
```
jaclyn-taroni, line 59: Unclear to me how I would ensure that I would have a copy of the current state file, but it may be the case that it has to be this way.

aws/setup-log.md
```diff
@@ -0,0 +1,63 @@
...
+With that file in the directory, you can test any changes with `terraform plan` and then implement them on AWS with `terraform apply`.
+
+Changes to the compute environments do not always seem to go smoothly, as Terraform does not always properly shut down the job queues.
```
jaclyn-taroni, line 62: 🆒

aws/setup-log.md
jaclyn-taroni: As I'm reading this, I am wondering about a local development section in the documentation, which almost certainly belongs in a different markdown file. This is a future problem.

aws/setup-log.md
jashapiro: Right now, you would email me, and I would send it to you. Kurt pointed me to terraform cloud, but I haven't looked into it yet.

aws/setup-log.md
jashapiro: That will need to come. But the hope is that most development won't require messing with terraform. Other AWS stuff will need to be added for the rest of the rep though, to get access to required buckets, etc.

PR #6, opened 2020-08-19
Add full transcriptome

jashapiro:
This PR adds the reference files and index for a full transcriptome, which is the union of the `cdna` and `ncrna` transcript files from Ensembl. This is created by concatenating the two relevant fasta files after download, which is then synced to S3

### DIFF
--- a/workflows/rnaseq-ref-index/README.md
+++ b/workflows/rnaseq-ref-index/README.md
@@ -6,13 +6,19 @@ The scripts in this directory implement a workflow for downloading and indexing 
 
 All reference files are downloaded from Ensembl for consistency, using the current version as of this writing: v100.
 Fasta files to be downloaded are listed in `fasta_ref_urls.txt` and other annotation files can be listed in `annotation_ref_urls.txt`.
-These files are then downloaded by running the bash script `get-refs.sh` which will download the selected files to a local machine, then sync to
-the following S3 location: `s3://nextflow-ccdl-data/reference/homo_sapiens/ensembl-100`
+These files are then downloaded by running the bash script `get-refs.sh` which will download the selected files to a local machine, then sync to the following S3 location: `s3://nextflow-ccdl-data/reference/homo_sapiens/ensembl-100`
+
+In addition, `get-refs.sh` will also combine the `cdna` and `ncrna` files into a single `txome` file to allow RNAseq mapping to all transcripts.
 
 ## Indexing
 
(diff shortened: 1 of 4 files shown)


workflows/rnaseq-ref-index/README.md
```diff
@@ -6,13 +6,19 @@ The scripts in this directory implement a workflow for downloading and indexing
...
 ## Indexing
 
 Currently, only salmon indexing is implemented, without decoys.
-The workflow expects to find the reference cdna file at `s3://nextflow-ccdl-data/reference/homo_sapiens/ensembl-100/fasta/Homo_sapiens.GRCh38.ncrna.fa.gz` and will place the salmon index directories at `s3://nextflow-ccdl-data/reference/homo_sapiens/ensembl-100/salmon_index` with subdirectories for each type of index.
+The workflow expects to find the following reference sequence files:
+
+- **cdna** at `s3://nextflow-ccdl-data/reference/homo_sapiens/ensembl-100/fasta/Homo_sapiens.GRCh38.cdna.all.fa.gz`
+- **txome** at `s3://nextflow-ccdl-data/reference/homo_sapiens/ensembl-100/fasta/Homo_sapiens.GRCh38.txome.fa.gz`
+
+and will place the salmon index directories at `s3://nextflow-ccdl-data/reference/homo_sapiens/ensembl-100/salmon_index` with subdirectories for each type of index.
```
jaclyn-taroni, line 21: I am almost certainly missing something (somethings?), but it is not clear to me why you would need both of these FASTA files here if `txome` is `cdna` + `ncrna`.

jashapiro: This workflow builds indexes for both the cdna alone and the full transcriptome. I could build the txome file on the fly, I suppose, but it seemed more straightforward to make it a separate file. Willing to reconsider though!

jaclyn-taroni: Ah, I see. I agree with that design decision. The disconnect was that I didn't understand that you are building two indices from the documentation. You do reference multiple types of index, but I was also looking at this in the context of the PR title, etc. - this is what I expected I was missing.

jashapiro: Yup, line 40-41 defines the reference files that will be used to build indexes, so both are built, though nextflow can (and did!) skip rebuilding references that had been previously built when run with the `-resume` flag.

PR #16, opened 2020-08-21
Add tx2gene and mitochondrial files

### LINKED ISSUE #10: create tx2gene files and mitochondrial gene lists
For both alevin and kallisto quantification, we need tx2gene files. 
These can be created from the gtf file downloaded as part of get-refs.sh, but it may be easiest to create them with the R {AnnotationHub} package, which can download the latest ensembldb objects. An example of doing this is here: https://github.com/AlexsLemonade/training-modules/blob/master/scRNA-seq/setup/mito_gene_lists.Rmd

It is probably also worth creating the mitochondrial gene lists that can be used as alevin inputs for its QC modules, which may be worth exploring. Even if we do not use those, we are likely to want to analyze mitochondrial percentages in the future.

jashapiro:
This PR adds a script to create and add tx2gene files using `AnnotationHub` and `ensembldb`.

Closes #10

At the moment, the version of Ensembl used is hard coded, but since that is true elsewhere in this repo at the moment, it should be fine. Making this more flexible is a future project.

This code does rely on the docker container defined in #15, which should be built and tagged `ccdl/scpca_r` before running the updated version of `get-refs.sh`. In practice though, this should not need to be run often, as once the files are created and synced to S3, they should not need to change unless there is a problem.

### DIFF
--- a/workflows/rnaseq-ref-index/build_tx2gene_mito.R
+++ b/workflows/rnaseq-ref-index/build_tx2gene_mito.R
@@ -1,0 +1,48 @@
+#!/usr/bin/env Rscript
+
+# This script uses Bioconductor tools to download ensembl annotation data,
+# create a transcript2gene mapping file for use in salmon alevin,
+# and a output a list of mitochondrial genes for use in QC.
+
+library(AnnotationHub)
+library(magrittr)
+library(optparse)
+
+# Set up optparse options
+option_list <- list(
+  make_option(
+    opt_str = c("-t", "--t2g_out"),
+    type = "character",
+    default = "annotation/Homo_sapiens.ensembl.100.tx2gene.txt",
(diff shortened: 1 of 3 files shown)


workflows/rnaseq-ref-index/build_tx2gene_mito.R
```diff
@@ -0,0 +1,48 @@
...
+library(AnnotationHub)
+library(magrittr)
+library(optparse)
+
+# Set up optparse options
+option_list <- list(
+  make_option(
+    opt_str = c("-t", "--t2g_out"),
+    type = "character",
+    default = "annotation/Homo_sapiens.ensembl.100.tx2gene.txt",
```
jaclyn-taroni, line 16: Any reason not to make this a TSV file?

Issue #9, opened 2020-08-20
Benchmark salmon/alevin and kallisto


jashapiro:
With indexes for salmon and kallisto made, we can start benchmarking the two on real data. We are interested in memory usage and speed, as well as concordence between the different mapping methods. 

For salmon, we should also look at the basic mapping vs decoy-aware mapping. I (and apparently developers of salmon) would prefer decoy-aware, but if it uses much more memory and/or time this may not be feasible.

jashapiro:
Initial results are presented in #18. Most interesting at first are the results in [trace.txt](https://github.com/AlexsLemonade/alsf-scpca/blob/11badd92f853b088b8597a83a75e2ee1e420aa0c/workflows/alevin-quant/trace.txt), which presents the memory and CPU usage of each process.

https://github.com/AlexsLemonade/alsf-scpca/blob/11badd92f853b088b8597a83a75e2ee1e420aa0c/workflows/alevin-quant/trace.txt#L1-L13

Notably, the full_sa jobs use much more memory, but do not seem to take much longer to run. However, the memory requirements are within the `m4.2xlarge` instance size when run with that instance's 8 cpus, making it seem likely that using the full selective alignment index is worth doing. 

I have not yet looked at mapping rate comparisons for kmer size cDNA vs transcriptome indexes or the number of ncRNA that appear in the samples. The full mapping results can be found at s3://nextflow-ccdl-results/scpca-benchmark/alevin-quant

jashapiro:
I have not gotten a formal analysis done, but I wanted to share some insights that I have gotten so far (some of which were in slack, but deserve posting here). I am analyzing the results in https://github.com/AlexsLemonade/alsf-scpca/blob/jashapiro/benchmark-analysis/workflows/alevin-quant/benchmark-analysis.Rmd but I have not yet filed a PR from that branch. 

The overall mapping rates for these samples seem low, in the 13-20% range for the two samples that I looked at. Looking at the mapped data though, things look pretty normal, with good numbers of mapped reads per cell and no over-abundance of mitochondrial reads. We did not do any kind of trimming on these data; I may give them a pass through `fastp` just to see if there is generally low quality sequence that gets removed that way and might explain the low mapping rate.

[...]

PR #18, opened 2020-09-01
Initial alevin benchmarking result

jashapiro:
This PR adds a workflow to test various alevin index types that could be used for mapping, notably using selective alignment and comparing full transcriptome and cdna only alignment. This fulfills some of the goals of #9. 

Also included is an html report for an initial run and a trace table describing memory and cpu usage. This was created by running the following command: 

```
nextflow -C ../nextflow.config run alevin-benchmark-indexes.nf -profile batch -with-report alevin-benchmark.html -with-trace
```

A quick look at these results seems to indicate that while full SA does take more memory, that memory usage is still within the m4.2xlarge range when run with 8 threads. No jobs died with OOM errors, (the one that did early on seems to have been a fluke).

All mapping results are stored at s3://nextflow-ccdl-results/scpca-benchmark/alevin-quant . Comparisons of the mapping results has not yet been performed.


jaclyn-taroni:
Jogging my own memory here - the decoy-aware indices are from pre-built from refgenie, is that correct?

jashapiro:
> Jogging my own memory here - the decoy-aware indices are from pre-built from refgenie, is that correct?

Yes, for this initial benchmarking, I used the prebuilt indices (indexes? I can't decide what I like here). That turns out to be Ensembl v97 with kmer 31.

envest:
Whoa, as you said, the two full selective alignment runs were definite memory usage outliers, but they obviously did fine within the testing environment allocation. I'm interested in the mapping comparison results to see if that memory usage is worth it 🤑  i.e. for my edification, when does SA make an important difference?

jashapiro:
> I'm interested in the mapping comparison results to see if that memory usage is worth it 🤑

Funny thing at the moment (maybe often, I don't know), an r4.2xlarge spot instance(~60GB ram) is cheaper than the standard m4.2xlarge (32GB RAM), by about 20%. I don't know what to do with this information.

As to selective alignment, I defer to the docs: https://salmon.readthedocs.io/en/latest/salmon.html and paper: https://www.biorxiv.org/content/10.1101/657874v2

workflows/alevin-quant/alevin-benchmark-indexes.nf
```diff
@@ -0,0 +1,72 @@
...
+  ch_reads = Channel.fromList(sample_ids)
+    // create tuple of [sample_id, [Read1 files], [Read2 files]]
+    .map{ id -> tuple("$id",
+                      file("${params.sample_dir}/${id}/*_R1_*.fastq.gz"),
+                      file("${params.sample_dir}/${id}/*_R2_*.fastq.gz"),
+                      )}
+  ch_indexes = Channel.fromList([
+    ['cdna_k31_no_sa',
+     's3://nextflow-ccdl-data/reference/homo_sapiens/ensembl-100/salmon_index/cdna_k31',
+     's3://nextflow-ccdl-data/reference/homo_sapiens/ensembl-100/annotation/Homo_sapiens.ensembl.100.tx2gene.tsv'],
```
jaclyn-taroni, line 48: The answer to this may be no - can you assign the tx2gene file as a variable?

workflows/alevin-quant/trace.txt.1
```diff
@@ -0,0 +1,13 @@
+task_id	hash	native_id	name	status	exit	submit	duration	realtime	%cpu	peak_rss	peak_vmem	rchar	wchar
```
jaclyn-taroni, line 1: I see you updated the trace and HTML files in https://github.com/AlexsLemonade/alsf-scpca/pull/18/commits/437842bf5c04fc3002bacd8a3ea02b4992ea8012 and this file is the "keep old trace for reference" based on the diff. I might suggest we name this something else for future us reasons. I would expect the other file to be the old one just looking at the file names. (This comment extends to the HTML files.)

jaclyn-taroni (approved):
I had a few questions, but no need for me to re-review.

workflows/alevin-quant/alevin-benchmark-indexes.nf
jashapiro: Not at this time, because there is still the partial SA in there from an external source, which does not use the same tx2gene. 😞

workflows/alevin-quant/trace.txt.1
jashapiro: Yeah, this was how `nextflow` handled the file name conflict. I would have expected the same as you did based on file names, but I am hesitant to rename them just for consistency. Scratch that, I will rename with a date.

Issue #4, opened 2020-08-19
Discussion: mapping and indexing options


jashapiro:
In working on #3, I have now reached the point where some decisions about mapping strategies going forward could be made. 

The current implementation of indexing (soon to be filed as a PR) creates a basic Salmon index. However, one of the major advantages of Salmon over kallisto is its support for selective alignment, so it seems that this would be worth implementing. 

There are two main strategies: partial decoys and full decoys. The latter is easier to create, the former creates a much smaller index and presumably allows running with less memory. Both kinds of indexes are available to download from refgenie, though figuring out exactly which annotation version was used can be a bit less than ideal, so there is some reason to implement creating our own decoy-based indexes (beyond just the practice writing workflows).

Notes for implementation:
Partial decoy index script described here: https://github.com/COMBINE-lab/SalmonTools/blob/master/README.md
Full decoy index procedure here:  https://combine-lab.github.io/alevin-tutorial/2019/selective-alignment/

What I can't find in those docs is discussion of the actual memory requirements for either creating these indexes or running salmon with them. I am sure such discussion exists, but I need to look it up.

jashapiro:
Another question is whether to map to just cDNA or the full transcriptome, including ncRNA as well. I have created indexes for the full transcriptome in #6. I would tend to vote for this if I don't hear an objection; it seems to me that ncRNA changes may well be worth looking at, but I am not sure what additional problems this might introduce.

jashapiro:
Adding to this index choice discussion, this twitter mini-thread: I asked the void about decoy sequences in Kallisto and got [this response](https://twitter.com/lpachter/status/1302780987764543489?s=20): 
> No. It's easy to implement but I'm not convinced it's a good idea, especially with human or mouse RNA-seq. When this was added to Salmon I tested it myself, running some controls, and I was unconvinced that it is actually improving results.

jashapiro:
Dropping an article in here about quantifying lncRNAs, which also seems to conclude that full transcriptome is worth doing to improve quantification of protein coding RNAs. https://academic.oup.com/gigascience/article/8/12/giz145/5663671

This, combined with preliminary results from benchmarking, has convinced me that the full transcriptome is the way to go, especially as lncRNAs may be important to some cancer processes. (see https://www.nature.com/articles/nrg3074/)

PR #20, opened 2020-09-08
Add full decoy transciptomes

jashapiro:
This PR adds full decoy transcriptome indices of the ensembl 100 reference for salmon to the index generation workflow. 

This should be a pretty straightforward addition, though whether these indices will actually be used in the future is still up for debate: #4


workflows/rnaseq-ref-index/build-index.nf
```diff
@@ -5,7 +5,8 @@ nextflow.enable.dsl=2
 params.ref_dir = 's3://nextflow-ccdl-data/reference/homo_sapiens/ensembl-100'
 params.cdna = 'fasta/Homo_sapiens.GRCh38.cdna.all.fa.gz'
 params.txome = 'fasta/Homo_sapiens.GRCh38.txome.fa.gz'
-params.kmer = [23, 31]
+params.genome = 'fasta/Homo_sapiens.GRCh38.dna.primary_assembly.fa.gz'
+params.kmer = '31'
```
jaclyn-taroni, line 9: I assume this change is intentional because only k = 31 full SA indexes are over on the benchmarking PR (#18). How does this change fit into the index generation big picture? Specifically, would this change how I would run `build-index.nf` if I still wanted to use k = 23 and I didn't have a local copy of those yet?

workflows/rnaseq-ref-index/build-index.nf
```diff
@@ -34,12 +36,40 @@ process salmon_index{
...
+    tuple val(index_base), path(reference), val(kmer)
+    path genome
+  output:
+    path "${index_base}_k${kmer}_full_sa"
+  script:
+    """
+    gunzip -c ${genome} \
+      |grep "^>" | cut -d " " -f 1 \
+      |sed -e 's/>//g' > decoys.txt
+    cat ${reference} ${genome} > gentrome.fa.gz
```
jaclyn-taroni, line 57: I've never built one of these before but these steps look consistent with [the tutorial](https://combine-lab.github.io/alevin-tutorial/2019/selective-alignment/).

jaclyn-taroni (approved):
👍  A question for my understanding and a comment for potential future reference.

workflows/rnaseq-ref-index/build-index.nf
jashapiro: This was intentional, yes, partly because initial benchmarking indicates that the k = 31 is better (faster, no lower mapping rate), but also because I found that `params` can't really be lists. Obviously, it works here, but there is no standard way to specify a list at the command line to override these parameters. 

To run with a different `params.kmer` value, you can run:
````
nextflow build-index.nf --kmer 23
```

Issue #19, opened 2020-09-01
Discussion: Universal Sample IDs and metadata table


jashapiro:
As samples come in, they are likely to be tagged and labelled in all kinds of different ways. We should maintain a single point of truth table with as much homogenized sample info as we might require. In particular, it will be useful to have a single format sample ID that we will be able to use across submitters.

I don't think we are currently requiring BioSample IDs from submitters, though that would seem to me to make sense if we had it. In the absence of that (or in addition) we should probably have an scpcaID of some sort. I am curious to hear what lessons learned from refine.bio could be applied here.

jashapiro:
After discussion with @kurtwheeler and @jaclyn-taroni, we have decided on the following scheme:

Samples will be identified by the following scheme: a SCPCS prefix (Single Cell Pediatric Cancer Sample) and six digit code, which we will increment as needed. This will will allow for up to 999,999 samples, which should be well in excess of our current needs (accommodating even this many would be well beyond the scope of the current project). 
- SCPCS000000

Since there may be more than one library for a given sample, we will also assign Run IDs, (modeled after SRA), which have similar format: a SCPCR prefix (Single Cell Pediatric Cancer Run) followed by six digits.
- SCPCR000000

Sample information will be stored in a `.tsv` file that is tracked in https://github.com/AlexsLemonade/ScPCA-admin, which will be mirrored to s3://ccdl-scpca-data/metadata/ via github actions. This will include information on tissue, collection dates, etc. as available. This will be compiled manually from submissions.

Library information will be stored in a separate `.tsv` file (stored and mirrored as above) that references the sample ID for each library, as well as the S3 prefix (folder) where the library files are stored, a list of library files and the checksum file provided for all files in the library. Filenames will be left as the submitter provided them to the extent possible, but each library will be placed at a separate S3 prefix (folder) named by the Library ID.

PR #21, opened 2020-09-09
Add kallisto workflow

jashapiro:
This PR adds a workflow for kallisto mapping. 

In most ways this is performing a similar function to the alevin mapping workflow, but as kallisto/bustools is more modular, I have embraced that and split the steps into separate processes: 1) pseudomapping with bustools; 2) correcting UMI assignments; 3) building gene counts.  

I initially wrote a process for creating a UMI whitelist file, which is still included here, but is unused. Instead, I am using the 10X genomics UMI barcode list as the whitelist. This is more conservative in some ways, as adjacent barcodes with very different counts might have been combined if the whitelist were constructed from the top barcodes, but in this case will only be corrected if a low count barcode does not exist in the original data set. 

Alevin, by contrast, does internal whitelist generation and correction; I am not sure if if has the 10X list internally or ignores this. Notably, the alevin option that allows providing a whitelist manually explicitly states that you should *not* provide the full UMI barcode list.

File structure, as before, will ultimately depend on #19 and related restructuring, but the general principle of separating the bus files from the counts I expect to be retained, though I am open to suggestions.


workflows/kallisto-quant/run-kallisto.nf
```diff
@@ -0,0 +1,131 @@
+#!/usr/bin/env nextflow
+nextflow.enable.dsl=2
+
+// run parameters
+params.ref_dir = 's3://nextflow-ccdl-data/reference/homo_sapiens/ensembl-100'
+params.index_dir = 'kallisto_index'
+params.index_name = 'cdna_k31'
+params.annotation_dir = 'annotation'
+params.t2g = 'Homo_sapiens.ensembl.100.tx2gene.tsv'
+params.mitolist = 'Homo_sapiens.ensembl.100.mitogenes.txt'
```
jaclyn-taroni, line 10: This doesn't get used anywhere yet, correct?

workflows/kallisto-quant/run-kallisto.nf
```diff
@@ -0,0 +1,131 @@
...
+  tag "${id}-${index}"
+  publishDir "${params.outdir}"
+  input:
+    tuple val(id), path(read1), path(read2)
+    path index
+  output:
+    path run_dir
+  script:
+    run_dir = "${id}-${index}"
+    // interleave read1 & read2 files
```
jaclyn-taroni, line 35: Tell me more?

workflows/kallisto-quant/run-kallisto.nf
```diff
@@ -0,0 +1,131 @@
...
+  input:
+    path run_dir
+    path tx2gene
+  output:
+    path count_dir
+  script:
+    count_dir ="${run_dir}/counts"
+    """
+    mkdir -p ${count_dir}
+    bustools count \
```
jaclyn-taroni, line 106: Because this is downstream of the UMI correction step, which includes sorting, the input here will be a sorted BUS file per the [docs](https://bustools.github.io/manual#count), correct?

jaclyn-taroni (approved):
:+1: just a few questions to check my understanding but nothing that precludes merging

workflows/kallisto-quant/run-kallisto.nf
jashapiro: It is not actually interleaving the data, just the file paths.
Took a while to figure this out, but this takes two lists, say [1, 2, 3] and [10, 20, 30], and makes the transpose matrix (if they had been combined): [[1,10], [2,20], [3,30]]. I then `flatten()` this to get the files in the order needed for kallisto, which expects R1 and R2 pairs in order. Finally `join()` with spaces between...

workflows/kallisto-quant/run-kallisto.nf
jashapiro: Correct.  It might never get used, I suppose.

workflows/kallisto-quant/run-kallisto.nf
jashapiro: That is correct. `output.corrected.bus` is also sorted, and will not exist unless the UMI correction step has run.

workflows/kallisto-quant/run-kallisto.nf
jaclyn-taroni: I might suggest

```suggestion
    // interleave read1 & read2 file paths
```

in that case because I definitely read it as interleaving the data!

PR #22, opened 2020-09-09
Initial benchmark analysis

jashapiro:
The benchmark analysis notebook here currently contains investigation of the count patterns seen in the cDNA vs. full transcriptome reference indexes, in service of the discussion in #4. I expect to add analysis of the decoy sequences soon (some parts of which can be seen here), but I thought it was worth getting this into a PR in its current state.

I initially tried to use some S3 libraries within R, but those did not seem to work the way I wanted, so I resorted to simply downloading the result files. to make that easier, I launch the docker image with the command described in README.md to import the local AWS tokens for access to S3 (the docker image includes the aws-cli tools). To run this notebook therefore requires that you already have run `aws setup` on your local system with credentials that allow read access to `s3://nextflow-ccdl-results`


jashapiro:
I have added some analysis of the decoy indexes to this notebook, which were more interesting than I expected! I am going to request @envest for review to start, but I think @cgreene and @jaclyn-taroni might want to have a look at the results as well.

Most intriguing results are right at the end, where I compare the full transcriptome to coding-only* indexes and notice that you get much more concordant results between the two when you include decoy sequences. Or I did something funny, which is always possible.

* Coding only includes pseudogenes, for some reason. Ask Ensembl.

The easiest way to view the notebook is probably http://htmlpreview.github.io/?https://github.com/AlexsLemonade/alsf-scpca/blob/jashapiro/benchmark-analysis/workflows/benchmarks/benchmark-analysis.nb.html#Comparing_decoy_indexes

envest (commented):
Overall I agree with the "go ahead with the full decoy" conclusion. For deeper understanding of the differences due to different indices, I think it may be worth including some additional samples, especially from another data contributor.

(Not taking sides on indexes vs. indices...) 😝

workflows/benchmarks/benchmark-analysis.Rmd
```diff
@@ -0,0 +1,324 @@
...
+)
+
+cor(compare_expression_full, method = "spearman")
+```
+```{r}
+ggplot(compare_expression_full, aes(log1p(cdna), y = log1p(txome))) +
+  geom_point()
+```
+
+Interestingly, the correlation between cDNA and txome expression values (for coding genes) using the full decoy seems substantially better than with the no decoy sequences!
```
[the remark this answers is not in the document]
jashapiro: It does!
No decoy:
image
Full decoy:
image

workflows/benchmarks/benchmark-analysis.Rmd
```diff
@@ -0,0 +1,324 @@
...
+)
+
+cor(compare_expression, method = "spearman")
+```
+```{r}
+ggplot(compare_expression, aes(x = log1p(cdna), y = log1p(txome))) +
+  geom_point()
+```
+Correlation is very good for genes present in both, with expression more often lower in the transcriptome set, which may make sense in the case of multimapping introduced by the larger number of potential targets.
+Interesting though that there are some genes which do not appear to be expressed in the cDNA set that have expression in the txome.
```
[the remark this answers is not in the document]
jashapiro: Pseudomapping is a bit of a black box?

workflows/benchmarks/benchmark-analysis.Rmd
```diff
@@ -0,0 +1,324 @@
...
+```
+ENSG00000269028, which is by far the most different, is the gene *MT-RNR2 like 12* which, aside from being related to at least 12 other genes, does not ring any particular bells.
+However, ENSG00000255823 is *MT-RNR2-like 8* 
+
+Other genes at the top of the differential list:
+ENSG00000163864: NMNAT3 nicotinamide nucleotide adenylyltransferase 3
+ENSG00000197563: PIGN Phosphatidylinositol Glycan Anchor Biosynthesis Class N
+ENSG00000173559: NABP1 Nucleic Acid Binding Protein 1
+
+I don't really know what to do with this list... it seems like few genes will have large differences, but a few differences are enormous!
```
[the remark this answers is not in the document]
jashapiro: Yeah, some of these seem consistent between the two samples, but I have no idea more generally.

workflows/benchmarks/benchmark-analysis.Rmd
```diff
@@ -0,0 +1,324 @@
...
+                  into = c("index_content", "kmer", "decoy"), 
+                  extra = "drop") %>%
+  dplyr::mutate(kmer = stringr::str_remove(kmer, "k"))
+  
+```
+
+### Get Annotations from AnnotationHub
+
+```{r}
+hub = AnnotationHub::AnnotationHub(ask = FALSE)
```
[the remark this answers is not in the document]
jashapiro: I've seen that occasionally, but I can't pin it down, so I am ignoring it. I don't love AnnotationHub (not the speediest), but it does seem to be nicely reliable.

workflows/benchmarks/benchmark-analysis.Rmd
```diff
@@ -0,0 +1,324 @@
...
+  dplyr::pull(gene_id) %>%
+  unique()
+# How many?
+length(detected_features)
+```
+
+How many of these are not protein coding?
+
+```{r}
+detected_nc <- detected_features[!detected_features %in% coding_genes]
```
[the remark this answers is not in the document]
jashapiro: Fair point for clarity. I am often confused by the operator precedence in situations like this.

workflows/benchmarks/README.md
```diff
@@ -0,0 +1,15 @@
+## Docker
+
+Running the benchmark analysis notebook should be done via docker, with the following command to mount the local directory as the home direcory and pass in the current user's AWS credentials and launch RStudio.
```
jaclyn-taroni, line 3:
```suggestion
Running the benchmark analysis notebook should be done via docker, with the following command to mount the local directory as the home directory and pass in the current user's AWS credentials and launch RStudio.
```

workflows/benchmarks/benchmark-analysis.Rmd
```diff
@@ -0,0 +1,324 @@
...
+```
+
+
+Combine all feature QC stats into a single data frame.
+```{r}
+sce_feature_qc <- purrr::map_df(sces, 
+                        ~ as.data.frame(rowData(.x)) %>%
+                          tibble::rownames_to_column(var = "gene_id"), 
+                        .id = "quant_id") %>%
+  dplyr::left_join(quant_info, by = c("quant_id" = "quant_dir"))
```
jaclyn-taroni, line 126: Very minor quibble - the spacing on these chunks seems _odd_ to me.

workflows/benchmarks/benchmark-analysis.Rmd
```diff
@@ -0,0 +1,324 @@
...
+quant_dirs <- list.dirs(data_dir, full.names = FALSE, recursive = FALSE)
+
+# split ids into components for later processing
+quant_info <- data.frame (quant_dir = quant_dirs, info = quant_dirs) %>%
+  tidyr::separate(info, sep = "[-]", 
+                  into = c("sample", "index_type")) %>%
+  tidyr::separate(index_type, 
+                  into = c("index_content", "kmer", "decoy"), 
+                  extra = "drop") %>%
+  dplyr::mutate(kmer = stringr::str_remove(kmer, "k"))
```
jaclyn-taroni, line 55: Nifty!

workflows/benchmarks/benchmark-analysis.Rmd
```diff
@@ -0,0 +1,324 @@
...
+```{r}
+cor(decoy_means[,3:5], method = "spearman", use = "complete.obs")
+```
+
+
+
+And we have them... interestingly, the best correlation is between no decoys and the partial decoy: the full decoy index is more different by this measure.
+This leads me to the preliminary conclusion that the partial index (which notably uses a slightly different set of transcripts at this stage, due to not being built locally) may not be worth pursuing, as it seems to make very little difference.
+
+Don't forget to look at your data!
```
jaclyn-taroni, line 281: 😂 💯

workflows/benchmarks/benchmark-analysis.Rmd
```diff
@@ -0,0 +1,324 @@
...
+cor(compare_expression_full, method = "spearman")
+```
+```{r}
+ggplot(compare_expression_full, aes(log1p(cdna), y = log1p(txome))) +
+  geom_point()
+```
+
+Interestingly, the correlation between cDNA and txome expression values (for coding genes) using the full decoy seems substantially better than with the no decoy sequences!
+
+This seems to me to be a good argument for using the full decoy, if it is less sensitive to the chosen transcript list.
```
jaclyn-taroni, line 324: Can we stick a `sessionInfo()` here? I know this should always be run in a container, but this practice has been helpful for tracking down potential weirdness in the past.

workflows/benchmarks/README.md
jaclyn-taroni: Can you add a brief reference to `.aws` like you have in `workflows/images/scpca-r/README.md` and/or to running `aws setup` ahead of time? Quoting you from your initial comment filing the PR:

> To run this notebook therefore requires that you already have run `aws setup` on your local system with credentials that allow read access to `s3://nextflow-ccdl-results`

jaclyn-taroni (approved):
This looks good to me in general! I had a couple documentation comments that should be addressed before merging, but no need for me to re-review before merging. 

I do agree with @envest's sentiment:

> For deeper understanding of the differences due to different indices, I think it may be worth including some additional samples, especially from another data contributor.

jashapiro:
> I do agree with @envest's sentiment:
> 
> > For deeper understanding of the differences due to different indices, I think it may be worth including some additional samples, especially from another data contributor.

When we have some, we can repeat?

jaclyn-taroni:
👍  works for me! Are the samples you've included so far the same cancer type? I was under the perhaps mistaken impression that there was more than one cancer type as part of this particular project and I'm curious if we'd see differences/less consistency across cancer types or if it would be like "weird transcript is weird" uniformly. This is a musing beyond the scope of this pull request.

jashapiro:
At this point, all the same type. We were supposed to get the other type, but there seems to be a delay...

PR #23, opened 2020-09-16
Kallisto processing example

jashapiro:
While we have examples from training for processing alevin files, and they are well-supported by `tximport`; we did not have similar results for processing kallisto|bustools files.  Kallisto also seems to do much less aggressive filtering of cells when creating the count matrix, so we have to do more on our own.

This notebook provides an example workflow for reading in the relevant result files after counting with bustools. I show three different methods for identifying cells. The current standard of using a knee plot is fast, so that seems likely to be the preferred method going forward. 

A more "principled" method is implemented in `DropletUtils` which I have also tested here. It is MUCH slower, but may be worth considering, if we think that recovering cells with low total expression may be important.


workflows/kallisto-quant/processing-kallisto.Rmd
```diff
@@ -0,0 +1,150 @@
...
+  subsets = list(mito = mito_genes[mito_genes %in% rownames(sce)])
+) %>%
+  scater::addPerFeatureQC()
+```
+
+```{r}
+ggplot(as.data.frame(colData(sce)),
+       aes(x = sum, y = detected, color = subsets_mito_percent)) + 
+  geom_point(alpha = 0.1) + 
+  scale_color_viridis_c()
```
jaclyn-taroni, line 126: I know `sum` and `detected` are `addPerCellQC()` choices and not Josh choices, but can you add more informative axis label here? Something like `detected genes` and `total count`.  I probably should have brought this up in the context of the alevin benchmarking as well. This is for future me that forgets what these values are.

workflows/kallisto-quant/processing-kallisto.Rmd
```diff
@@ -0,0 +1,150 @@
...
+sce_ed <- scater::addPerCellQC(
+  sce_ed, 
+  subsets = list(mito = mito_genes[mito_genes %in% rownames(sce_ed)])
+) %>%
+  scater::addPerFeatureQC()
+
+ggplot(as.data.frame(colData(sce_ed)),
+       aes(x = sum, y = detected, color = subsets_mito_percent)) + 
+  geom_point(alpha = 0.1) + 
+  scale_color_viridis_c()
```
jaclyn-taroni, line 146: Same comment about axis labeling.

workflows/kallisto-quant/processing-kallisto.Rmd
```diff
@@ -0,0 +1,150 @@
...
+ggplot(as.data.frame(colData(sce)),
+       aes(x = sum, y = detected, color = subsets_mito_percent)) + 
+  geom_point(alpha = 0.1) + 
+  scale_color_viridis_c()
+```
+
+
+### EmptyDrops QC
+
+Out of curiousity, what do the EmptyDrops results look like?
```
jaclyn-taroni, line 132: This seems more permissive to me? Specifically, there's a bunch of "cells" with low mitochondrial %, but also very few total counts and detected genes.

workflows/kallisto-quant/processing-kallisto.Rmd
```diff
@@ -0,0 +1,150 @@
...
+## Make SCE object 
+
+```{r}
+cells <- which(rank_df$total >= knee)
+sce <- SingleCellExperiment(list(counts = counts[ , cells]))
+```
+
+That results in a much more reasonably sized object, one we can work with as usual with SCE objects. 
+
+## Simple QC
```
jaclyn-taroni, line 111: What does the simple QC entail? Is `addPerFeatureQC()` making some additional choices by default to do filtering or is this in reference to the knee method?

workflows/kallisto-quant/processing-kallisto.Rmd
```diff
@@ -0,0 +1,150 @@
...
+```
+
+### Knee plot method
+
+Find the knee, keep the cells above that rank.
+
+```{r}
+system.time(rank_df <- DropletUtils::barcodeRanks(counts))
+knee = metadata(rank_df)$knee
+inflection = metadata(rank_df)$inflection
```
jaclyn-taroni, line 71:
```suggestion
system.time(rank_df <- DropletUtils::barcodeRanks(counts))
knee <- metadata(rank_df)$knee
inflection <- metadata(rank_df)$inflection
```

workflows/kallisto-quant/processing-kallisto.Rmd
```diff
@@ -0,0 +1,150 @@
...
+is.cell <- drop_df$FDR <= 0.01 
+sum(is.cell, na.rm=TRUE)
+```
+
+This method also keeps many more "cells", which may be net negative for downstream processing as well, but may also be worth pursuing for some kinds of analysis.
+
+For now, we will proceed with the knee method to define cells.
+
+## Make SCE object 
+
```
jaclyn-taroni, line 103:
```suggestion
## Make SCE object 

For now, we will proceed with the knee method to define cells.

```

This is where I expected to find that information as a reader, not under the header about EmptyDrops.

jaclyn-taroni (commented):
This is looking good! I had a few comments in service of helping 3-months-from-now-Jaclyn (that is to say documentation/labeling comments).

workflows/kallisto-quant/processing-kallisto.Rmd
jashapiro: Yes, it is much more permissive, though it claims to be principled about it: keeping drops with few counts if they are different from "the bulk". I would tend to expect that it is doing too little filtering, and some minimum count would still be warranted, but where that would be is an open question.

workflows/kallisto-quant/processing-kallisto.Rmd
jashapiro: No additional choices... I just meant that I was making some simple QC plots!

workflows/kallisto-quant/processing-kallisto.Rmd
jaclyn-taroni: Is there a reason `=` needs to be used for assignment here? I didn't think that deeply about it; I had a knee jerk reaction.

jaclyn-taroni (approved):
👍  I had one outstanding question about assignment, but nothing that prevents approval.

PR #24, opened 2020-09-22
WIP: compare alevin & kallisto quantification

jashapiro:
Here I have started a comparison of Alevin and Kallisto results for the first batch of samples

A few note and takeaways: 

- I am using the knee plots to determine Kallisto cells included. This results in many fewer cells included than whatever metric Alevin is using internally (I am taking it's list as given).

- Mapping rates seem higher with kallisto and mean expression of genes tends to be higher in kallisto.
  - Kallisto UMI/cell tends to be higher than with Alevin, as is transcripts detected/cell.

- The correlation between methods is not all that great. Spearman rho <0.8? I am troubled by this, but don't quite know what to do to make a decision between the two at this point. See image below.

plot


cgreene:
I'm not really sure spearman is the best metric for something like single cell data where you're likely to have a lot of low counts where a count being assigned to one transcript instead of another might make a massive difference in ranks. What does the linear correlation look like for comparison's sake?

jashapiro:
Yes, it is a bit unfair, but I also didn't want to weight by the highly expressed outliers! Pearson gives r = .98-.99, or if I take `log(n+1)` (which isn't right either) ~0.9

cgreene:
Ok - it sounds like it's not _that_ different. Those highly expressed genes are probably also the ones measured with the least error. If it was just a variance funnel I think that'd make a lot of sense given the nature of the data. It does look like there might be a bit of systematic difference with kallisto routinely giving higher expression values to certain lowly expressed genes (it looks like there might be one distribution that's a bit of a hockey stick and another distribution that more or less lies along the diagonal).

jashapiro:
Kallisto's mapping rate *appears* higher though it is a bit hard to judge; it is unclear if they report the same calculation for mapping rate (i.e. one might be reporting of all reads, the other of reads with valid barcodes). But if we assume kallisto is more tolerant in its mapping, we might get something like that pattern? Those extra mappings will be distributed among some subset of genes more subject to crossmapping, making the floor higher? I'm speculating here.

jashapiro:
I am closing this PR in favor of a new PR #35  that also includes Cellranger.

PR #30, opened 2020-10-07
Use metadata to get individual run info

### LINKED ISSUE #28: Update workflows to use SCPCR ids and locations
The workflows to generate count matrixes (alevin and kallisto) will need to be updated for new file locations and ID patterns. 

The new inputs will be a set of SCPCR ids and a metadata file (`scpca-library-metadata.tsv`) which contains the file locations to be used.

This update may also add checking  md5 hashes of the raw files. 

Output files should designate runs by their SCPCR id.

jashapiro:
This PR updates the nextflow workflows to work from the metadata files that we have created to track sample information. 

Closes #28 

The general procedure is to read in the metadata file and use the nextflow function `.splitCsv()` to process that file into fields that we can then access when creating the channel which will be used by the rest of the workflow. 

I also added a workflow (in `checks`) that can be used to check the md5 hashes of uploaded files. It currently checks all files from all runs in the metadata file. This seems reasonable, as long as we don't run it often. I think it is safe to assume that integrity checks from S3 are sufficient for most usage, and this can be used just to verify that the files we have are the ones that we expect to have been sent. 

### For reviewers:

I would recommend starting with `check-md5.nf`, as in many ways that is the simplest workflow here, and because it is new, the whole file is easily visible in the github interface. 

While the code is not long, there are a number of things going on here that may be unfamiliar to new nextflow users. One is the use of operator chaining (the `.function()` syntax in the workflow) which is very much like piping. Another is the use of closures `{ }` for functions like `map`. There is also at least one place where a variable appears to be used before it is defined.

All of this means that I fully expect questions on how things are actually working. Don't hesitate to ask!


workflows/checks/check-md5.nf
```diff
@@ -0,0 +1,29 @@
...
+  container 'ubuntu:20.04'
+  publishDir "${params.outdir}"
+  input:
+    tuple val(id), val(md5_file), path(files)
+  output:
+    path outfile
+  script:
+    outfile = "${id}-md5check.txt"
+    """
+    md5sum -c ${md5_file} > ${outfile}
```
cbethell, line 18: Here is the step where the md5checksum is ran and output assigned for each file, correct?

workflows/checks/check-md5.nf
```diff
@@ -0,0 +1,29 @@
+#!/usr/bin/env nextflow
+nextflow.enable.dsl=2
+
+// run parameters
+params.run_metafile = 's3://ccdl-scpca-data/sample_info/scpca-library-metadata.tsv'
+params.outdir = 's3://nextflow-ccdl-results/scpca-checks/'
+
+process check_md5{
```
cbethell, line 8: Ah so a process is like a custom function essentially.

workflows/alevin-quant/run-alevin.nf
```diff
@@ -48,12 +49,14 @@ process alevin{
 }
 
 workflow{
-  sample_ids = params.sample_ids?.tokenize(',') ?: []
-  ch_reads = Channel.fromList(sample_ids)
+  run_ids = params.run_ids?.tokenize(',') ?: []
```
cbethell, line 52: Is there where the `id` value come from in `check-md5.nf`?

workflows/alevin-quant/run-alevin.nf
```diff
@@ -48,12 +49,14 @@ process alevin{
...
+  ch_reads = Channel.fromPath(params.run_metafile)
+    .splitCsv(header: true, sep: '\t')
+    .filter{it.scpca_run_id in run_ids} // use only the rows in the sample list
```
cbethell, line 55: This filtering here is like saying filter `i in run_ids`?

workflows/alevin-quant/run-alevin.nf
```diff
@@ -48,12 +49,14 @@ process alevin{
...
     // create tuple of [sample_id, [Read1 files], [Read2 files]]
-    .map{ id -> tuple("$id",
-                      file("${params.sample_dir}/${id}/*_R1_*.fastq.gz"),
-                      file("${params.sample_dir}/${id}/*_R2_*.fastq.gz"),
+    .map{row -> tuple(row.scpca_run_id,
```
cbethell, line 57: If so, does this mean that this is also acting as `i`, allowing iteration on rows in the metadata file.

workflows/checks/check-md5.nf
```diff
@@ -0,0 +1,29 @@
+#!/usr/bin/env nextflow
+nextflow.enable.dsl=2
+
+// run parameters
+params.run_metafile = 's3://ccdl-scpca-data/sample_info/scpca-library-metadata.tsv'
+params.outdir = 's3://nextflow-ccdl-results/scpca-checks/'
+
```
cbethell, line 7: You did not define the params like you did in the other scripts (i.e `params.run_ids = "SCPCR000001,SCPCR000002"`), are they being passed into this script somewhere? Why are they not needed here?

cbethell (commented):
Lots of questions below @jashapiro! Along with one general question and that is:

Do any of these scripts overlap or are sourced within another? It's not obvious to me that any of them are if they are.

workflows/checks/check-md5.nf
```diff
@@ -0,0 +1,29 @@
+#!/usr/bin/env nextflow
+nextflow.enable.dsl=2
```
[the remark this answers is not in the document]
jashapiro: Good question. The answer is that nextflow syntax changed relatively recently and this tells nextflow we are going to use the new syntax. 

https://www.nextflow.io/docs/latest/dsl2.html

workflows/checks/check-md5.nf
```diff
@@ -0,0 +1,29 @@
...
+
+// run parameters
+params.run_metafile = 's3://ccdl-scpca-data/sample_info/scpca-library-metadata.tsv'
+params.outdir = 's3://nextflow-ccdl-results/scpca-checks/'
+
+process check_md5{
+  container 'ubuntu:20.04'
+  publishDir "${params.outdir}"
+  input:
+    tuple val(id), val(md5_file), path(files)
```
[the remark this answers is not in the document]
jashapiro: Sort of... this is a super strange syntax, I think, partly because the inclusion/exclusion of parenthesis seems strange.

This is more like the definition of parameters for a function. In `R`, this might look like
```
function(id, md5_file, files){}
```
But it is a bit stranger than that, because really the first argument expects a tuple, so it is more like:
```
function(my_tuple){
id = my_tuple[1]
md5_file = my_tuple[2]
files = my_tuple[3]
...
}
```

In other words, this is actually defining where we will put the values for `id`, `md5_file` and `files`. The actual assignment comes when we call the process later with `check_md5(ch_runs)`.  So `ch_runs` will pass a tuple with three values, which will then get assigned to `id`, `md5_file` and `files`.

`id` and `md5_file` will be assigned as a `val` variable type (strings in this case), where `files` is a `path`, which lets nextflow know that there are files associated with it. We need to tell nextflow that so it knows that it will need to get those files to run the process. 

(An odd thing here is that `md5_file` is also a file in fact, so you would think it needs to be gotten too, but it is one of the files in `files` and we aren't allowed to tell nextflow to get the same file twice.)

workflows/checks/check-md5.nf
```diff
@@ -0,0 +1,29 @@
...
+  script:
+    outfile = "${id}-md5check.txt"
+    """
+    md5sum -c ${md5_file} > ${outfile}
+    """
+}
+workflow{
+  ch_runs = Channel.fromPath(params.run_metafile)
+    .splitCsv(header: true, sep: '\t')
+    .map{row -> tuple(row.scpca_run_id,
```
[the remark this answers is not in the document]
jashapiro: Yes, they are functions. The dot is acting like a pipe: they are working on the output of the previous line.

(Technically they are "methods" of the "object" that precedes them, but the above explanation is simpler)

workflows/checks/check-md5.nf
```diff
@@ -0,0 +1,29 @@
+#!/usr/bin/env nextflow
+nextflow.enable.dsl=2
+
+// run parameters
+params.run_metafile = 's3://ccdl-scpca-data/sample_info/scpca-library-metadata.tsv'
+params.outdir = 's3://nextflow-ccdl-results/scpca-checks/'
```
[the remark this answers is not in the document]
jashapiro: Yes, they are setting the paths; Nextflow will get or put the files as needed.

Making them `params` means we will be able to change them at runtime from the command line if we want to.

workflows/checks/check-md5.nf
```diff
@@ -0,0 +1,29 @@
...
+    """
+}
+workflow{
+  ch_runs = Channel.fromPath(params.run_metafile)
+    .splitCsv(header: true, sep: '\t')
+    .map{row -> tuple(row.scpca_run_id,
+                      row.md5_file,
+                      file("s3://${row.s3_prefix}/*")
+                      )}
+  check_md5(ch_runs)
```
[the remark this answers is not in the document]
jashapiro: Yeah, that is pretty much right. We define the processes we want above, with what inputs and outputs they expect. Then we call them here, giving a "channel" (or more) as the input.

Channels are *the thing* with nextflow. In some ways they are just a list, but when they go to a process, the process pulls one element at a time. So first we run `check_md5()` with the first element of the channel, then we run it again on the second, then the third, until we reach the end. So in this case, each element of the channel is a tuple, with `id`, `md5_file` and `files` in it. So something like ["SCPR000001", "md5file.txt", [file1, file2, file3]] (`files` is actually a tuple of its own). 

So the whole process is something like this: we make a channel of a file (so it starts as a list with one thing in it). We then use `.splitCsv()` to parse through that file and make a channel with one element for each line in the file. (`. splitCsv()` does work on tsv files if we define the `sep` as a tab, which we did.). Then each of those row elements gets passed to the `.map{}` function, which takes each row element and structures it into a new tuple with just the things we need. That output is what we will then give to `check_md5()`.  Does that make sense? It is kind of convoluted in a way, but it starts to make sense after you wrap your head around it, I hope.

workflows/checks/check-md5.nf
jashapiro: Yes, but a bit more. These are the steps that will run independently. So each process has to be able to run on a separate machine, and it defines all the files and resources it needs.

workflows/alevin-quant/run-alevin.nf
jashapiro: Sort of... this is a super strange thing, actually, and I got it straight from the developer in a chat conversation.

`params.run_ids` is the string defined at the top (or it can be given at the command line). So in this case by default it is "SCPCR000001,SCPCR000001"

`.tokenize(',')` is basically splitting at the comma and turning it into a list (like `str_split(',')`).  

The strange part is the `?` What that does is say that if `params.run_ids` doesn't exist or is blank, then skip the `tokenize` step and just return an empty list. 

I'm then using that list to filter the data table in the metadata file to only the samples I want to actually run.

workflows/alevin-quant/run-alevin.nf
jashapiro: Yeah, kind of. `it` is a special variable that is kind of like `.` in dplyr. It is the default name for each element passed to a function like `.filter{}` which will run on one element of the channel/list at a time.

If I wanted to rename it (which I do later) I could have used 
```
.filter{row -> row.scpca_run_id in run_ids}
```
Why do one or the other?  ¯\_(ツ)_/¯

workflows/alevin-quant/run-alevin.nf
jashapiro: Yep! I could have used `it` here too, but row makes more sense?

workflows/checks/check-md5.nf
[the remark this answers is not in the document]
cbethell: Ahh interesting, it does make much more sense now! So essentially once nextflow gets the files in `files`, each file in `files` becomes a `val` variable type.

workflows/checks/check-md5.nf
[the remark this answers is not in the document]
cbethell: Neat (thank you for the simpler and _technical_ explanation, but most importantly for the simpler first)!

workflows/checks/check-md5.nf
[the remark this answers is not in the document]
cbethell: Your explanation was extremely helpful, this make so much more sense in my head now! Everything is now actually connecting, thank you 💯

workflows/checks/check-md5.nf
cbethell: Gotchaa!

workflows/alevin-quant/run-alevin.nf
cbethell: Ah okay, this also makes more sense to me now. I agree the `?` part is strange, but makes sense when you think about it ¯\_(ツ)_/¯

workflows/alevin-quant/run-alevin.nf
cbethell: Right! okay gotcha 👍

cbethell (approved):
LGTM and equally or possibly more importantly, what's in this PR makes a whole lot more sense to me! 🎉

workflows/checks/check-md5.nf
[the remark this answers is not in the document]
jashapiro: Sort of? I think they stay a list of `paths`, but functionally we can mostly treat them the same. There is some weirdness where the actual location of the file in question is not something we are supposed to care about. I ran into this when I made assumptions that turned out to be incorrect, but I don't remember the details.

jashapiro:
> LGTM and equally or possibly more importantly, what's in this PR makes a whole lot more sense to me! 🎉

That is the goal! At some point, you may have to edit or write one of these super-strange workflows!

PR #31, opened 2020-10-08
Add cellranger index script

jashapiro:
This PR adds a nextflow workflow for generating a cellranger index file. 

The workflow currently builds two versions of the index: one with just protein coding genes, the other with all regions defined in the ensembl GTF. I have considered filtering the second, using the annotations similar to the ones defined in the [cellranger reference](https://support.10xgenomics.com/single-cell-gene-expression/software/pipelines/latest/advanced/references#mkgtf), but stuck with the simplest versions in this case.

This PR includes a Dockerfile to create an image with cellranger installed, and a readme with instructions for building that image and uploading it to AWS ECR for use with AWS batch/nextflow. Due to license restrictions, we can't make this public, leading to the use of ECR. I wanted to put it in the github container registry, but couldn't figure out how to make that work nicely with nextflow/AWS, so ECR it is (it works smoothly, just with ugly container ids!).

### For reviewers:

I would like to check that the dockerfile can be downloaded by other users in our organization. You should be able to login with the instructions in the readme and pull the current image. Let me know if you run into any troubles.

The "correct" filtering of the gtf file is another question I would like to think about, though I think that is probably worthy of a separate issue. Feel free to put thoughts in your review, and I will reference them in a new issue if/when it comes.

Any general questions on the nextflow format are also welcome!


workflows/images/cellranger/README.md
```diff
@@ -0,0 +1,36 @@
+This folder contains a Dockerfile for the cellranger analysis.
+
+The cellranger archive component must be downloaded separately to comply with licensing, and should be placed in this folder.
+The current version of this file is `cellranger-4.0.0.tar.gz` and can be downloaded from [10X Genomics Website](https://support.10xgenomics.com/single-cell-gene-expression/software/downloads/latest) after agreeing to their license terms.
```
cansavvy, line 4: This being here made me think I should download this first, but I do not need to after I read past this. 

Some headers to separate where the "instruction steps" begin would be helpful.

workflows/images/cellranger/README.md
```diff
@@ -0,0 +1,36 @@
...
+
+
+```
+docker build . -t scpca-cellranger:4.0.0
```
cansavvy, line 8: This command doesn't work in the top of the repository (and this branch) as I expected it to. I get this warning: `unable to prepare context: unable to evaluate symlinks in Dockerfile path:` are there symlinks you've set that you were supposed to have worked? Or was I not supposed to run this command at the top of the repository?

workflows/images/cellranger/README.md
```diff
@@ -0,0 +1,36 @@
...
+```
+
+This is then tagged and pushed to or pulled from AWS ECR with the following steps:
+
+First, login to the AWS ECR docker repository with:
```
cansavvy, line 13: To confirm: this next command should be run with the docker image I just built?

cansavvy (commented):
Following the PR prompt I tried to follow the README instructions, before I go further and make this work otherwise I have a couple questions about how you intended me to run these steps.

cansavvy:
Oh and you may want to add `.DS_Store` to the .gitignore

workflows/images/cellranger/README.md
jashapiro: You do need to download this if you want to build the docker image. But if you have access to our image stored in github, you don't need it. I will clarify.

workflows/images/cellranger/README.md
jashapiro: This will not work unless you have downloaded cellranger, I think. There should not be any symlinks, but I wonder if that is the error you get if a file is missing.

jashapiro:
> Oh and you may want to add `.DS_Store` to the .gitignore

Maybe, but I have it in my system-wide .gitignore, which I highly recommend!
https://gist.github.com/linuslundahl/577073

jashapiro:
Updated the docker image readme to try to add some clarity. Let me know if that helps, @cansavvy

cansavvy:
> > Oh and you may want to add `.DS_Store` to the .gitignore
> 
> Maybe, but I have it in my system-wide .gitignore, which I highly recommend!
> https://gist.github.com/linuslundahl/577073

Oh but then that only works for you and not other people who use this repo.

workflows/images/cellranger/README.md
```diff
@@ -0,0 +1,39 @@
+This folder contains a Dockerfile for the cellranger analysis.
+
+## Building the image
+
+In order to build this image, the cellranger archive component must be downloaded separately to comply with licensing, and should be placed in this folder.
```
cansavvy, line 5: By `this folder` you mean `workflows/images/cellranger`?

cansavvy: I assume this should also be my working directory for the next command.

workflows/images/cellranger/README.md
```diff
@@ -0,0 +1,39 @@
...
+At this point, the image should be ready for use on the local machine.
+
+## Pushing to AWS ECR
+If the updated image needs to be pushed to AWS ECR, you can follow the steps below to push to the Alex's Lemonade ECR repository.
+
+First, login to the AWS ECR docker repository with:
+```
+aws ecr get-login-password --region us-east-1 | docker login --username AWS --password-stdin 589864003899.dkr.ecr.us-east-1.amazonaws.com
+```
+Note that for the code above to work, you must have set up `aws` command line tools on your machine, and have run [`aws configure`](https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-quickstart.html) to set up your credentials for AWS access.
```
cansavvy, line 22: We don't want to run this in a docker image? I don't have aws installed so I'll have to do this. Seems like this would still be handy to have in a docker image too, but it seems you have a lot of docker images going on?

cansavvy: Does it matter if I get version 1 or version 2? https://docs.aws.amazon.com/cli/latest/userguide/install-cliv2.html

workflows/images/cellranger/README.md
```diff
@@ -0,0 +1,39 @@
...
+This is a private repository; access is only available to Alex's Lemonade (CCDL) users at this time.
+
+Following login (which only needs to be done once), tag and push the docker image with:
+
+```
+docker tag scpca-cellranger:4.0.0 589864003899.dkr.ecr.us-east-1.amazonaws.com/scpca-cellranger:latest
+docker tag scpca-cellranger:4.0.0 589864003899.dkr.ecr.us-east-1.amazonaws.com/scpca-cellranger:4.0.0
+docker push 589864003899.dkr.ecr.us-east-1.amazonaws.com/scpca-cellranger:4.0.0
+docker push 589864003899.dkr.ecr.us-east-1.amazonaws.com/scpca-cellranger:latest
```
cansavvy, line 31: I think I pushed a docker image successfully! 

Next question is do we prefer tagging with `:4.0.0` or `:latest`? Might be easier to choose one.

workflows/images/cellranger/README.md
jashapiro: No. This is infrastructure that most people don't need to set up or reproduce. Not doing dockers inside dockers for people to use AWS.

workflows/images/cellranger/README.md
jashapiro: We want both; if we update cellranger, we still want something tagged with `latest`. And don't push over my image! (Hopefully you got a message that the image was already present?)

workflows/images/cellranger/README.md
jashapiro: Yes, the folder where this file lives, and yes, this should be the working directory, will clarify.

workflows/images/cellranger/README.md
cansavvy: It said some layers were already present, but it still seems like it pushed something. I don't have the log message anymore.

jashapiro: I have clarified that this is not something people should do in most cases... Switched order to emphasize `pull` over `push`.

workflows/images/cellranger/README.md
cansavvy: Does it matter if its version 1 or 2 though?

cansavvy (approved):
I just have my one last question, but otherwise, this LGTM!

workflows/images/cellranger/README.md
jashapiro: I don't think it does. I apparently have v1, which is what was available through `conda`.
Some investigation shows that `aws ecr get-login-password` is not available on versions below 1.17, but that is pretty old.

PR #33, opened 2020-10-12
Use Launch Templates for AWS Batch configuration

### LINKED ISSUE #14: Explore Launch Templates
Currently the compute environments on AWS batch are defined by an AMI, but it is likely that this AMI does not have sufficient disk space for all tasks. To accommodate jobs with larger requirements, we could create a larger AMI, but this requires maintaining a snapshot with the larger file, most of which is blank space. It should be possible to use launch templates to create jobs with arbitrary attached disks, which would allow flexibility on launch, rather than at AMI creation time.

One question with this is that the current AMI, the docker base size is set in `/etc/sysconfig/docker-storage` with the option `--storage-opt dm.basesize=50GB` to match the image size. For a larger image, this would presumably need to be reset. 

It may be possible to set up docker to use a different volume setup on invocation to get around this limitation. It should be possible to set docker flags within nextflow, and these settings should be storable in the config file to attach only to jobs with specific labels, but this is an area I have not yet explored.

jashapiro:
This PR does three main things in service of making our AWS batch settings more flexible for future use (and allowing running cellranger in the near future).

Closes #14

First, it switches the AMI used to one based on Amazon Linux 2, which seems to make use of a more flexible docker setup with regard to the space available to the docker image.

Second, it uses Launch Templates to allow the same AMI to be used to launch instances with varying disk sizes. There are two templates currently implemented: one with a 64GB EBS volume, the other with a 500GB volume.

Third, these Launch templates are now used to create the compute environments and batch queues. I have replaced the "priority" queue with a "bigdisk" queue.

I have tested these changes by submitting previous jobs to the default spot queue, and it seems to work! I have not explicitly checked the bigdisk queue, but I see no reason it should not also work as expected. It will be tested with the cellranger workflow that is coming soon, and I expect any changes required to be implemented at that point.


aws/nextflow-compute.tf
```diff
@@ -44,39 +46,80 @@ resource "aws_batch_compute_environment" "nf_spot" {
...
+    }
     # ec2_key_pair = aws_key_pair.nf_keypair.key_name
     security_group_ids = [
       aws_security_group.nf_security.id,
     ]
     subnets = [
       aws_subnet.nf_subnet.id,
     ]
-    type = "EC2"
+    type = "SPOT"
```
jaclyn-taroni, line 74: With the `bid_percentage = 100` change this is effectively the same thing but with greater flexibility for the future (i.e., you can decrease the `bid_percentage`)?

jaclyn-taroni: Your documentation answered this question so that's a good sign 😄

jaclyn-taroni (approved):
👍 LGTM with the caveat that I don't know much about terraform and we will find out shortly about any changes that need to be made!

PR #35, opened 2020-10-15
Compare mappers and pseudomappers

jashapiro:
The notebook here is an update to the one originally filed as part of #24, adding in Cellranger as a "baseline" comparison.

⚠️ Depends on #30 and #31, so those should be merged first, which should eliminate some of the extraneous file changes.

The top line result is that the results from Alevin are much more in line with the Cellranger baseline than Kallisto's results. There is a subtle further improvement with the addition of decoys, but it is small. 

As discussed earlier (https://github.com/AlexsLemonade/alsf-scpca/pull/24#issuecomment-696928398) Kallisto seems to consistently give somewhat higher expression values for lowly expressed genes; Alevin does this too relative to Cellranger (to a lesser extent), which leads me to suspect that this is a result of pseudomapping, which may be partially offset by Alevin's pseudomapping. 

Overall, I would be happy to make the decision to use Alevin for future mapping: it is much faster and less resource-hungry than Cellranger, and its results are impressively well correlated, even using the somewhat unfair standard of a Spearman correlation. 

Note: while this analysis was done using cDNA transcripts only, I have no reason to suspect (partly based on results from #24) that using the full transcriptome would change the results at all. Indeed, #22 suggests that the full transcriptome is better, but I only ran Cellranger with cDNA, so I was limited to that comparison for this notebook.


jashapiro:
Hmm... a lot of stuff snuck in here. Reviewers should focus on `quant-comparisons.Rmd`, which can be viewed in it's current form here:
https://htmlpreview.github.io/?https://github.com/AlexsLemonade/alsf-scpca/blob/jashapiro/compare-mappers/workflows/benchmarks/quant-comparisons.nb.html

workflows/benchmarks/quant-comparisons.Rmd
```diff
@@ -0,0 +1,524 @@
+---
+title: "SCRNAseq quantification comparisons"
+author: "Joshua Shapiro for CCDL"
+output: 
+  html_notebook:
+    toc: true
+    toc_float: true
+---
+
+This notebook contains analysis of some initial benchmarking in runs of Alevin and Kallisto
```
jaclyn-taroni, line 10: Probably needs a refresh with the CellRanger addition 😁

workflows/benchmarks/quant-comparisons.Rmd
```diff
@@ -0,0 +1,524 @@
...
+```
+
+Summarize the number of detected genes per cell.
+```{r}
+summary(cellranger_cell_qc$detected)
+```
+
+
+
+## Comparing summary stats across tools
```
jaclyn-taroni, line 303: This seems like a good point to break this into it's own notebook. And if it were me, I'd probably write a few functions! 3 tools = do things 3x = time to write function.

jaclyn-taroni (commented):
These analyses seem totally reasonable! The content in `workflows/benchmarks/quant-comparisons.Rmd` seems like two notebooks to me: 1. get the output of the tools and add QC to your SingleCellExperiment objects 2. dig in and make some plots. 

I would also raise the documentation level overall–imagine a reader doesn't remember what's up with a kallisto knee plot, but they're maybe interested in adapting `read_kallisto_counts()` to get started. I envision that these benchmark notebooks _will be user facing_ because if **_I_** was using the summarized data, I'd want to know why we picked what we picked. It seems like we are in a good position right now to get these in that kind of shape.

My _personal preference_ would be to write a bunch of functions once it's time to plot because there are 3 tools here, but doesn't rise to "hill I am willing to die on" levels by any means.

jashapiro:
Yes, I agree with all this! I wanted to get *something* in to justify future directions, but it does all need some time to dry and polish. 

Some of the repetitiveness is also a function of process: I was adapting what started as a 2 mapper comparison, so adding *just one more* seemed not too bad. But yeah, functions would be good.

I was also thinking this notebook maybe should move out of `workflows` as this particular analysis doesn't introduce or depend on a particular workflow. It is more directly about the mappers and not quite "benchmarking" in the sense of speed and resources.

Side note: I need to file an issue on overall level of documentation in this repo. I think things are getting solid enough that usage needs to be written down more than I have done so far... see also https://github.com/AlexsLemonade/alsf-scpca/pull/34#discussion_r509551764. I think I had hesitated to write too much while things were very much in flux, but the time has come.

jaclyn-taroni:
Procedurely, I think we should plan to have this PR essentially serve as a _draft_ and put in new PRs intended for merging  that are more polished as you suggest (bonus: changes will not have snuck in). That series of PRs would also cover the restructuring of `workflows` you mentioned in my mind.

Side note: if we're happy with the Alevin decision, I think it's probably time to file some bulk workflow tickets! 🎟️

jashapiro:
Yes, I was about to convert this to a draft before going to work on the parts...

jashapiro:
closing as it has now been replaced by #39 and #45.

PR #38, opened 2020-10-29
Make the main readme informative

### LINKED ISSUE #37: Update top level README.md
It is time for the main README for this project to be a real README!

I envision this should describe the overall structure of the repository, including the contents of the `aws` and `workflow` directories, as well as where the data files are stored. 

General usage instructions, including how to access to AWS and run a nextflow pipeline, should be described briefly, with links to more detailed instructions. Some of these will be external sites, but also pointers to any other READMEs within the repo that cover more specific types of analyses and tools.

jashapiro:
In reviews of some of the specific workflows, it became apparent that this repo would benefit greatly from some basic instructions and description of the way files interact. This PR attempts to address this need with an expansion of the main README.md file.

Closes #37 

## For reviewers

Please read through these instructions and let me know if you think they are sufficiently detailed, or perhaps too detailed? I was not aiming for a complete description of every step, but one that would be sufficient for most internal users, with links to documentation for details about each step rather than duplicating primary sources as much as possible. 

In general, my philosophy would prefer people read the primary docs and learn about the tools in general rather than just relying the examples in the document. That said, I don't want people to be stuck on the basics they need to use the workflows with the most common options.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,6 +9,7 @@ Management and analysis tools for ALSF Single-cell Pediatric Cancer Atlas data.
   - [AWS](#aws)
     - [Installing AWS Command line tools](#installing-aws-command-line-tools)
     - [Configuring your AWS credentials](#configuring-your-aws-credentials)
+    - [AWS infrastructure](#aws-infrastructure)
 - [Running Workflows](#running-workflows)
 
 ## Environment Setup
@@ -63,6 +64,10 @@ You may want to set up your `Default region name`: most of the computing resourc
 The `Default output format` can be left as `None`
 
 The `aws configure` command will create a directory at `~/.aws` with a `config` file and a `credentials` file that will be required for `nextflow` commands to work smoothly with Batch and S3.
+
+#### AWS infrastructure
+
+The infrastructure on AWS (Batch queues, AMIs, etc.) is defined via [terraform](https://www.terraform.io) using the files in the `aws` directory. More details are described in [aws/setup-log.md](aws/setup-log.md)
 
 
 
--- a/README.md
+++ b/README.md
@@ -11,6 +11,8 @@ Management and analysis tools for ALSF Single-cell Pediatric Cancer Atlas data.
     - [Configuring your AWS credentials](#configuring-your-aws-credentials)
     - [AWS infrastructure](#aws-infrastructure)
 - [Running Workflows](#running-workflows)
+- [Running locally](#running-locally)
+  - [Running on AWS Batch](#running-on-aws-batch)
 
 ## Environment Setup
 
@@ -79,6 +81,10 @@ This will allow easy access to the shared `nextflow.config` file and keep all in
 In particular, `nextflow` creates a separate work directory for every task: these will appear by default (for local tasks) in the `workflows/work` directory if the command is invoked from `workflows`.
 As the work directories can get large, it is helpful to have a single location to keep track of and purge as needed.
 
+## Running locally
+
+⚠️ Running workflows in this repository locally is not something to take lightly! It is fine for a quick test, and useful for development, but note that most of the workflows here use very large data files, which will have to be downloaded locally to run. Workflow processing may require large amounts of RAM and time, so the following example commands will rarely be used, and are mostly for illustration. ⚠️
+
 The basic command to run a workflow will look something like the following:
 
 ```
@@ -93,7 +99,11 @@ In most cases, you will want to skip any cached steps that have already run: thi
 nextflow run alevin-quant/run-alevin.nf -resume
 ```
 
-To run the same workflow on AWS Batch, make sure your credentials are configured [as described above](#configuring-your-aws-credentials), and then run with the `batch` profile.
+⚠️ Again, you probably don't want to run locally unless you have a good reason and know the limitations! ⚠️
+
+### Running on AWS Batch
+
+To run the same workflow on AWS Batch, make sure your credentials are configured [as described above](#configuring-your-aws-credentials), and then run with the `batch` profile, which has been configured in `nextflow.config` for the CCDL infrastructure.
 
 ```
 nextflow run alevin-quant/run-alevin.nf -profile batch -resume
--- a/README.md
+++ b/README.md
@@ -83,12 +83,16 @@ As the work directories can get large, it is helpful to have a single location t
 
 ## Running locally
 
-⚠️ Running workflows in this repository locally is not something to take lightly! It is fine for a quick test, and useful for development, but note that most of the workflows here use very large data files, which will have to be downloaded locally to run. Workflow processing may require large amounts of RAM and time, so the following example commands will rarely be used, and are mostly for illustration. ⚠️
+⚠️⚠️⚠️
+Running workflows in this repository locally is not something to take lightly!
+It is fine for a quick test, and useful for development, but note that most of the workflows here use very large data files, which will have to be downloaded locally to run.
+Workflow processing may require large amounts of RAM and time, so the following example commands will rarely be used, and are mostly for illustration.
+⚠️⚠️⚠️
 
 The basic command to run a workflow will look something like the following:
 
 ```
-nextflow run alevin-quant/run-alevin.nf
+nextflow run checks/check-md5.nf
 ```
 
 This will run the workflow locally, using the default parameters as defined in the workflow file.
@@ -96,7 +100,7 @@ This will run the workflow locally, using the default parameters as defined in t
 In most cases, you will want to skip any cached steps that have already run: this can be done by adding the `-resume` flag.
 
 ```
-nextflow run alevin-quant/run-alevin.nf -resume
+nextflow run checks/check-md5.nf -resume
 ```
 
 ⚠️ Again, you probably don't want to run locally unless you have a good reason and know the limitations! ⚠️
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -1,2 +1,114 @@
...
+
+Optionally, you can track your nextflow workflows in progress with [Nextflow Tower](https://tower.nf/).
+To use this, you will need to create an account on https://tower.nf/ (using your github account is likely simplest), then get your token value from https://tower.nf/tokens.
+You will then want to create an environment variable with this token value:
+`export TOWER_ACCESS_TOKEN=<MYTOWERTOKEN>`
+This line can be entered at the command line, but you will likely want to add it to your shell initialization file (`~/.bash_profile`, `~/.zshrc`, etc.)
+
+### Docker
+
+All of the workflows are designed to run via Docker containers for each tool. It is therefore critical that you have Docker installed for any local execution. For downloads and installation, go to https://www.docker.com/products/docker-desktop
```
cansavvy, line 40: But the Docker does not have Nextflow or AWS?

jashapiro: Hmm, I think one of them does, but that is irrelevant here. Most of the docker tools are single-use, as in they are designed to run just one tool. Nextflow runs a whole bunch of different ones, but that is mostly automatic. You don't need to (and should not) build or pull any docker images manually to run anything.

cansavvy: Gotcha. Okay. So if I install Nextflow and AWS I should be good to go.

cansavvy: In this case, can you move this toward the end of the document or as a subnote underneath Nextflow?  This way I don't accidentally think I'll have to install this too (even though I realize you have no "Install Docker" section).

jashapiro: You *do* have to install Docker. "It is therefore critical that you have Docker installed for any local execution." But you just have to have the engine installed, and that is best covered by the linked docs.

cansavvy: OMG. Right.  Never mind.

README.md
```diff
@@ -1,2 +1,124 @@
...
+The workflows for this repository are stored in the `workflows` directory, organized by task into subdirectories.
+In general, workflows should be run from the main `workflows` directory, rather than the subdirectories.
+This will allow easy access to the shared `nextflow.config` file and keep all intermediate and log files in a single location.
+In particular, `nextflow` creates a separate work directory for every task: these will appear by default (for local tasks) in the `workflows/work` directory if the command is invoked from `workflows`.
+As the work directories can get large, it is helpful to have a single location to keep track of and purge as needed.
+
+## Running locally
+
+⚠️ Running workflows in this repository locally is not something to take lightly! It is fine for a quick test, and useful for development, but note that most of the workflows here use very large data files, which will have to be downloaded locally to run. Workflow processing may require large amounts of RAM and time, so the following example commands will rarely be used, and are mostly for illustration. ⚠️
+
```
cansavvy, line 87: Nextflow told me: 
`Caused by:
  Process requirement exceed available memory -- req: 28 GB; avail: 16 GB` I don't know how it knows 28 GB is what it needs, but you could put that down somewhere (if that is accurate).

README.md
```diff
@@ -1,2 +1,124 @@
...
+  - [AWS](#aws)
+    - [Installing AWS Command line tools](#installing-aws-command-line-tools)
+    - [Configuring your AWS credentials](#configuring-your-aws-credentials)
+    - [AWS infrastructure](#aws-infrastructure)
+- [Running Workflows](#running-workflows)
+- [Running locally](#running-locally)
+  - [Running on AWS Batch](#running-on-aws-batch)
+
+## Environment Setup
+
```
cansavvy, line 18: Could put a note at the top about this probably won't run locally. (Not sure if I needed to do this environment set up locally since it won't be able to run?)

README.md
```diff
@@ -1,2 +1,124 @@
...
+In most cases, you will want to skip any cached steps that have already run: this can be done by adding the `-resume` flag.
+
+```
+nextflow run alevin-quant/run-alevin.nf -resume
+```
+
+⚠️ Again, you probably don't want to run locally unless you have a good reason and know the limitations! ⚠️
+
+### Running on AWS Batch
```
cansavvy, line 104: If this is more likely how this will be run, you could list this first before the `Run locally` section which probably won't work (which you do say many times that it probably won't).

README.md
jashapiro: You do need to do the environment setup regardless; it needs to have the aws credentials to run on batch. Docker is not needed if you are not running locally, but you need everything else.

README.md
jashapiro: hmm.. yeah, that's what I get for not testing instructions. I could give an example that will run locally, but all the mapping workflows request more processors... I think that is something I should fix at the config level. I want to have a local example, as I use them often for testing, and I wanted to build up the important nextflow options sequentially.

jashapiro: Okay, fixed the workflow to only request lots of memory and cpu if running in batch. It should work now, restoring the state I meant to have where running locally deserved a caution, but should not outright fail (except cellranger, which will fail because it is cellranger).

cansavvy: I'll try running again then.

README.md
```diff
@@ -1,2 +1,124 @@
...
+
+The workflows for this repository are stored in the `workflows` directory, organized by task into subdirectories.
+In general, workflows should be run from the main `workflows` directory, rather than the subdirectories.
+This will allow easy access to the shared `nextflow.config` file and keep all intermediate and log files in a single location.
+In particular, `nextflow` creates a separate work directory for every task: these will appear by default (for local tasks) in the `workflows/work` directory if the command is invoked from `workflows`.
+As the work directories can get large, it is helpful to have a single location to keep track of and purge as needed.
+
+## Running locally
+
+⚠️ Running workflows in this repository locally is not something to take lightly! It is fine for a quick test, and useful for development, but note that most of the workflows here use very large data files, which will have to be downloaded locally to run. Workflow processing may require large amounts of RAM and time, so the following example commands will rarely be used, and are mostly for illustration. ⚠️
```
cansavvy, line 86: How do I download the files for a local run?

cansavvy: Less significant note, I think there are three sentences on one line here.

cansavvy: Or should download have already happened when I ran Nextflow? Right now, Alevin doesn't like the file: 
```
Exception : [Error reading from the FASTA/Q stream. Minimum return code for left and right read was (-2). Make sure the file is valid.]
  salmon alevin was invoked improperly.
  For usage information, try salmon alevin --help
  Exiting.
```

cansavvy: It looks like it should run from the s3 bucket link? I double checked my aws config, all seems fine -- I would have expected an AWS warning/error if not being able to access the files was the problem.

jashapiro: Nextflow handles all the file downloads itself.

Yeah, I got that error too... I think it means something went wrong with the download? I hadn't seen it until today, and it doesn't seem to have recurred when I deleted the work directory and reran the workflow (though it has yet to complete 🤞🏼 )

jashapiro:
I just switched the examples to the check-mdf5 workflow, which should be simpler. Still requires big file downloads, but the processing is faster (still not fast, because big files).

README.md
```diff
@@ -1,2 +1,132 @@
...
+The `aws configure` command will create a directory at `~/.aws` with a `config` file and a `credentials` file that will be required for `nextflow` commands to work smoothly with Batch and S3.
+
+#### AWS infrastructure
+
+The infrastructure on AWS (Batch queues, AMIs, etc.) is defined via [terraform](https://www.terraform.io) using the files in the `aws` directory. More details are described in [aws/setup-log.md](aws/setup-log.md)
+
+
+
+## Running Workflows
+
```
cbethell, line 77: I believe it may also be beneficial to have a list or table of the available workflows (or even just the tasks that the workflows in each subdirectory accomplish) in this main `README.md`. What do you think?

README.md
```diff
@@ -1,2 +1,132 @@
...
+The workflows for this repository are stored in the `workflows` directory, organized by task into subdirectories.
+In general, workflows should be run from the main `workflows` directory, rather than the subdirectories.
```
cbethell, line 79: Ah I missed this line the first time around when testing and wondered why I kept getting the error `Remote resource not found: https://api.github.com/repos/nextflow-io/checks/contents/check-md5.nf`.

cbethell (commented):
The `README` looks great and was extremely helpful! I just left a suggestion and a small question below re where files are stored when running locally vs when running on AWS.

README.md
```diff
@@ -1,2 +1,128 @@
...
+This will allow easy access to the shared `nextflow.config` file and keep all intermediate and log files in a single location.
+In particular, `nextflow` creates a separate work directory for every task: these will appear by default (for local tasks) in the `workflows/work` directory if the command is invoked from `workflows`.
```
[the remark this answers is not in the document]
cbethell: Ah well I see you just answered the first part of my question in the last commit.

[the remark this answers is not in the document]
jashapiro: I didn't actually address final output files as this is workflow specific: I should note that fact.

README.md
jashapiro: At this point, I don't think this is needed: we are still very much in flux! If I were to do this, I think the appropriate place would be `workflows/README.md`

README.md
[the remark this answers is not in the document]
jashapiro: added in 8dc9bc43d545f01742b0891ba55ae848c6760e9d

README.md
cbethell: Two very good points!

README.md
```diff
@@ -1,2 +1,135 @@
...
+⚠️⚠️⚠️
+Running workflows in this repository locally is not something to take lightly!
+It is fine for a quick test, and useful for development, but note that most of the workflows here use very large data files, which will have to be downloaded locally to run.
+Workflow processing may require large amounts of RAM and time, so the following example commands will rarely be used, and are mostly for illustration.
+⚠️⚠️⚠️
+
+The basic command to run a workflow will look something like the following:
+
+```
+nextflow run checks/check-md5.nf
```
cansavvy, line 98: I like this example as opposed to the old one. It seems to have run successfully. It took my computer 1h 7m. If that's a normal run time, may want to put a note that this does take awhile to run at first.

README.md
cansavvy: I think a table would be nice too, when the time comes. Should we put this on an issue?

cansavvy (approved):
Okay. This seems ready to me!

cbethell (approved):
LGTM 🚀

PR #53, opened 2021-01-15
Make the md5 process a bit more robust, and print failures

jashapiro:
This PR has a couple of changes to deal with the cases which pop up when the disk fills up downloading files to check md5s (I would like to not download the files at all, but that would require a more extensive rewrite). 

The other thing that this adds is a set of final output checks that will print any md5 mismatches to stdout when the workflow is run.  This can save the user from having to scan the output file manually to identify any failures.

### DIFF
--- a/workflows/checks/check-md5.nf
+++ b/workflows/checks/check-md5.nf
@@ -27,9 +27,24 @@ process check_md5{
     outfile = "${id}-md5check.txt"
     """
     md5sum -c ${md5_file} > ${outfile}
-    cat ${outfile}
     """
 }
+
+process cat_md5{
+  // combine md5 results files, removing blank lines
+  container 'ubuntu:20.04'
+  publishDir "${params.outdir}"
+  input:
+    path(files)
+  output:
+    path outfile
+  script:
+    outfile = "00_all_md5check.txt"
+    """
+    cat ${files} | sed '/^\\s+\$/d' > ${outfile}
+    """
+}
+
 workflow{
   run_ids = params.run_ids?.tokenize(',') ?: []
   run_all = run_ids[0] == "All"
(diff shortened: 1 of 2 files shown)


workflows/checks/check-md5.nf
```diff
@@ -13,6 +13,11 @@ params.run_ids = "SCPCR000001,SCPCR000002"
 process check_md5{
   container 'ubuntu:20.04'
   label 'bigdisk'
+  // tasks sometimes fail due to disk space but we can't specify that directly
+  errorStrategy 'retry' 
+  maxRetries = 1
+  cpus { 1 + 3 * (task.attempt - 1)  } //use more CPUs to take over a machine if failure
```
cbethell, line 19: Is `task.attempt` initialized elsewhere? If so, where? If not, can you explain why it works here without being defined or initialized elsewhere?

Also, if I understand the above equation correctly, does this mean that the process will only be retried once if needed, and on that _second_ try, we will use (1 + 3) times more CPUs than (task.attempt - 1) which would be (2 -1). Is that correct?

cbethell, line 19: Is `task.attempt` initialized elsewhere? If so, where? If not, can you explain why it works here without being defined or initialized elsewhere?

Also, if I understand the above equation correctly, does this mean that the process will only be retried once if needed, and on that _second_ try, we will use (1 + 3) times more CPUs than (task.attempt - 1) which would be (2 -1). Is that correct?

cbethell (approved):
LGTM! I have one or two questions below but otherwise the updates in this PR were pretty straightforward to review and check out!

workflows/checks/check-md5.nf
jashapiro: `task.attempt` is provided by Nextflow, and reflects task attempt count, so it is just 1, 2, 3, etc. (though since we only retry once, we can't get `task.attempt` > 2). The idea is that the first time it will use 1 cpu, the second time it will use 4, which is usually a full AWS machine, but we don't directly control that, so it may not always work 🤷🏼

cbethell: Ahh gotcha! 👍

Issue #55, opened 2021-03-17
Create docker image for Cellranger 6.0


allyhawkins:
Cellranger has recently been updated to 6.0. The updated cellranger includes use of STARsolo, making alignment faster for scRNA-seq and a new potential option for pre-processing of ScPCA data. In order to run cellranger 6.0 with samples from ScPCA, we need to create a new docker image that uses cellranger 6.0 (instead of the current 4.0 that is used), and then test cellranger 6.0 on a subset of the samples received so far using the nextflow cellranger-quant script. 

In addition to creating the new docker image, we also will check if we need to remake the STAR index for 6.0. If so, a new issue will be created. Updates will also need to be made to the nextflow cellranger-quant script if new arguments are needed for cellranger 6.0. 

[Updates about cellranger 6.0](https://support.10xgenomics.com/single-cell-gene-expression/software/pipelines/latest/release-notes)
[Notes on STARsolo release](https://github.com/alexdobin/STAR/blob/master/RELEASEnotes.md)

PR #56, opened 2021-03-18
New docker image for cellranger 6.0.0

allyhawkins:
This PR addresses #55 which requires addition of a new docker image for cellranger 6.0.0 to be used for benchmarking pre-processing of ScPCA data. 

The old dockerfile used for cellranger 4.0.0 was moved to a new file named `cellranger4.docker` and the new `Dockerfile` contains the image for running cellranger 6.0.0. The docker image was built and tested using the following command: `docker run -it --rm scpca-cellranger:6.0.0 cellranger testrun --id=tiny` to ensure that it successfully can run cellranger 6.0.0. 

Additionally, it appears that the reference files for cellranger have not been updated since July 7, 2020 and the reference files located on S3 were added in October, 2020. The STARsolo website also doesn't mention any need for a new reference index. I think it is safe to assume that we do not need to update the reference. https://support.10xgenomics.com/single-cell-gene-expression/software/downloads/latest

Things that have not been done: 
- this updated image has not been pushed to AWS ECR
- the nextflow workflow `workflows/cellranger-quant/run-cellranger.nf` has not been updated yet and will be addressed in a new issue

### DIFF
--- a/images/cellranger/Dockerfile
+++ b/images/cellranger/Dockerfile
@@ -2,8 +2,8 @@ FROM ubuntu:20.04
 LABEL maintainer="ccdl@alexslemonade.org"
 
 WORKDIR /opt
-COPY cellranger-4.0.0.tar.gz cellranger-4.0.0.tar.gz
-RUN tar xzf cellranger-4.0.0.tar.gz && rm cellranger-4.0.0.tar.gz
-ENV PATH="/opt/cellranger-4.0.0:${PATH}"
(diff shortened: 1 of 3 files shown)


jashapiro (approved):
Looks good! On second thought, I am not sure we need to keep the old version of the Dockerfile around, but it won't hurt either. We can always remove it later.

jashapiro:
Actually... I realized that there is one more step that needs to be done... updating `images/cellranger/README.md` to reflect the changes. You can either do that in this PR as well before merging, or in a separate PR.

Issue #52, opened 2021-01-04
Add flexibility for 10Xv2 data


jashapiro:
Right now, Alevin and kallisto workflows are hard coded to use 10Xv3, but it looks like some of the data we will be getting is likely to be 10Xv2. Workflows will need to be modified to take the 10X version from the metadata file and apply the correct settings in the Alevin workflow.

PR #65, opened 2021-04-06
Add 10Xv2 support to kallisto workflow

jashapiro:
This PR adds a bit more flexibility to the kallisto workflow, starting work on https://github.com/AlexsLemonade/alsf-scpca/issues/52.  The main thing is that it now supports 10Xv2 and 10Xv3.1. It does this by now having a map (dictionary) of the 10X barcode files that can be checked when correcting cell barcodes as well as filtering to technology types that are in that list. 

I checked that the workflow ran as expected with SCPCR000119 (v3.1) & SCPCR000120 (v2), though I have not looked at the actual count results. 

In the process of  testing, I learned that `bustools count` will segfault if given the results from a `bustools correct` step where the barcode file is not found. You'd think `bustools correct` or the following `bustools sort` would fail, but they don't.

I also did a bit of cleanup, removing some unnecessary publish steps. I am still publishing the full `bus` directory, but only once now.

### DIFF
--- a/workflows/kallisto-quant/run-kallisto.nf
+++ b/workflows/kallisto-quant/run-kallisto.nf
@@ -127,7 +127,7 @@ workflow{
   run_all = run_ids[0] == "All"
   samples_ch = Channel.fromPath(params.run_metafile)
     .splitCsv(header: true, sep: '\t')
-    .filter{it.technology == "10Xv3"} // only 10X data
+    .filter{it.technology in barcodes} // only 10X data
     // use only the rows in the sample list
     .filter{run_all || (it.scpca_run_id in run_ids)}
   // create tuple of [sample_id, technology, [Read1 files], [Read2 files]]
--- a/workflows/kallisto-quant/run-kallisto.nf
+++ b/workflows/kallisto-quant/run-kallisto.nf
@@ -28,7 +28,6 @@ process kallisto_bus{
   container 'quay.io/biocontainers/kallisto:0.46.2--h4f7b962_1'
   label 'cpus_8'
   tag "${id}-${index}"
-  publishDir "${params.outdir}"
   input:
     tuple val(id), val(tech), path(read1), path(read2)
     path index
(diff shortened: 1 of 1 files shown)


allyhawkins (approved):
All looks good and seems to run without issue. I just had a few questions for my own clarification and understanding.

You said in your comments that you changed it to only publish the bus directory once but it appears that the `publishDir` is still in every process? Did you still mean for that to be the case?

Also the `bustools_whitelist ` process still exists but it's never used. Is this because of what you were referring to in terms of causing a segfault if it can't find the barcodes file? It looks like it should always be using one of the provided 10x files to me so theoretically we could remove that process. I tested it with removing that process with SCPCR000004 and it worked so was wondering if there is another reason why we needed it?

jashapiro:
> You said in your comments that you changed it to only publish the bus directory once but it appears that the `publishDir` is still in every process? Did you still mean for that to be the case?

No, I meant to take those out. I had an accidental loss of some edits at one point, and I guess I forgot to put those deletions back in.  Removing those for later.
 
> Also the `bustools_whitelist ` process still exists but it's never used. Is this because of what you were referring to in terms of causing a segfault if it can't find the barcodes file? It looks like it should always be using one of the provided 10x files to me so theoretically we could remove that process. I tested it with removing that process with SCPCR000004 and it worked so was wondering if there is another reason why we needed it?

That was there when I thought we might want it to act a bit more like `alevin` and generate its own barcode list from the data. But this was not recommended, so I took out that step from the workflow, leaving the process definition in case we wanted to add it back at some point. We could remove it, but the cost is so small to leave it...

allyhawkins:
> That was there when I thought we might want it to act a bit more like `alevin` and generate its own barcode list from the data. But this was not recommended, so I took out that step from the workflow, leaving the process definition in case we wanted to add it back at some point. We could remove it, but the cost is so small to leave it...

Understandable to leave it then. I just wanted to make sure I understood why it was there, since it wasn't necessary for the workflow to actually run, but perhaps maybe useful in the future.

PR #66, opened 2021-04-06
Make md5 check disk efficient

jashapiro:
The `check-md5.nf` workflow was failing periodically due to staging too many large files, and the ways I was trying to get around the problem were not working all the time, leading to intermittent failures.

Here I rewrote the workflow to use a python script that gets the files from S3 directly, streaming them to calculate the md5 hash without the files ever hitting the disk. This also allows a bit more customization of the output over `md5sum`.

I had to make a new container for this, and while I was at it I installed the aws cli in the container, even though we probably don't need that any time soon.


allyhawkins (approved):
LGTM. Definitely not a boto3 expert with python, but it looks like this is doing what you're asking and checking that md5 files are consistent with the fastq files present in S3. Everything seems to work when run and show that everything is where it should be.

PR #67, opened 2021-04-06
Add tags to terraform resources

jashapiro:
This PR adds CCDL-standard default tags to the AWS compute resources we are using. I have already applied these changes to the aws batch setup, and it seems to be fine.

This is a first step in addressing https://github.com/AlexsLemonade/ScPCA-admin/issues/122


jaclyn-taroni (approved):
Not a terraform expert, but the tags LGTM!

Issue #54, opened 2021-03-17
Add premature RNA reference


jaclyn-taroni:
@jashapiro commented on [Tue Jan 26 2021](https://github.com/AlexsLemonade/ScPCA-admin/issues/105)

In reading #81 and other project descriptions, it struck me that we may want to add a premature RNA reference for some of the single cell, and particularly single nucleotide samples. This may allow us to improve mapping rates, and potentially add some RNA velocity analysis down the line. 

An example workflow can be found with the alevin tutorials website: https://combine-lab.github.io/alevin-tutorial/2020/alevin-velocity/

PR #68, opened 2021-04-06
Adjust reference scripts to make pre mRNA fasta needed to generate index

allyhawkins:
These files include updates needed to generate a pre-mRNA index for single-nuclei RNA sequencing with alevin and kallisto, see #54. Before actually making the index, we need to create an expanded fasta file that includes the sequences for unspliced and spliced mRNA to use to build the index. 

I followed the steps in the Alevin tutorial [here](https://combine-lab.github.io/alevin-tutorial/2020/alevin-velocity/) and after some reading of this [preprint](https://www.biorxiv.org/content/10.1101/2020.03.13.990069v1.full.pdf) it appears that we should be able to use this pre mRNA fasta to generate both a salmon and kallisto index. 

In order to do this, I needed to update the dockerfile for `images/scpca-r` which I have done and it is currently able to be pulled from github using `docker pull ghcr.io/alexslemonade/scpca_r:v2`.  I also updated the reference files that are being used to the most recent release - ensembl-103 and have created a new directory in  `s3://nextflow-ccdl-data/reference/homo_sapiens/ensembl-103`. I'm not entirely sure how necessary this is, but while I was working on this I went ahead and did this. 

The script to actually generate the fasta file, `make_pre_mrna_fasta.R` currently runs on it own and generates the necessary output files, but when run inside the container, it stops mid script, failing in the middle of the `eisaR::getFeatureRanges` call without any error... I'm going to continue to try and troubleshoot this, but also in the meantime, I will make sure that the nextflow processes are updated for actually building the mRNA index using the output since having the indexes built is the ultimate goal. It does appear to me that this is an error with running the script in the container and not an error with the script itself since it is able to run independently of the container, unless I'm missing something.


workflows/rnaseq-ref-index/make_pre_mrna_fasta.R
```diff
@@ -0,0 +1,96 @@
...
+
+# write the associated annotations to gtf 
+eisaR::exportToGtf(
+  grl, 
+  filepath = expanded_gtf
+)
+
+# create text file mapping transcript and intron identifiers to 
+# corresponding gene identifiers
+df <- eisaR::getTx2Gene(
```
jashapiro, line 94: Can we use this to generate the Tx2Gene for the main index too, or do we want to generate _all_ of the transcript references with this script?

jashapiro (commented):
Partial comments here... I don't yet know why it might be failing in Docker, there seems to be something going on there for me too, but it might be Mac/Docker version specific?

images/scpca-r/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM rocker/tidyverse:4.0.2
+FROM rocker/tidyverse:4.0.5
```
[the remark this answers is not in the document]
allyhawkins: So I tried 4.0.4 and that still threw errors with matrixStats being too old, and then was still seeing the same thing with 4.0.5. So to refer to your below comment about adding in matrixStats, it was still needed even with using the most up to date R image. I did not however try it with 4.0.4 and matrixStats being installed separately if we do want a more stable version. But like you said, I'm not sure how important that is considering running in the docker image is not working anyways.

images/scpca-r/Dockerfile
```diff
@@ -17,7 +17,18 @@ RUN apt-get update -qq && apt-get -y --no-install-recommends install \
...
+
 #### R packages
+###############
+# this comes first since bioconductor packages are 
+# dependent on updated matrixStats
+RUN Rscript -e "install.packages(c( \
+    'here', \
+    'optparse', \
+    'rprojroot', \
+    'matrixStats') \
```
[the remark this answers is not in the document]
allyhawkins: One would think, but it was still failing with the error that matrixStats did not have the most up to date version. That error is gone when installing it separately.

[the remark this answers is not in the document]
jashapiro: When I built locally, this did not seem to be a problem. I could remove this line and everything worked. Not sure what is going on, tbh.

workflows/rnaseq-ref-index/make_pre_mrna_fasta.R
allyhawkins: So I was looking into this and theoretically yes we could grab the spliced and unspliced genomic regions and then generate the Tx2Gene mapping with `eisaR` for the main index. My concern with doing this, which is potentially minor, but in the previous script, you had grabbed the versioned transcript ID's. The way that the `eisaR::getTx2Gene` works is that it converts from a GenomicRanges object. Currently using the `eisaR::getFeatureRanges` function it doesn't look like I can grab versioned transcript ID's but only get spliced vs. unspliced. So I'm not sure if we would want that for our main index? Additionally, we would still have to use AnnotationHub to extract the mito genes for QC. 

So I'm thinking I will just modify the scripts to get the regions corresponding to unspliced transcripts rather than intronic regions and then merge `build_tx2gene_mito.R` and `make_pre_mrna_fasta.R` into one script that for now will be run outside of a docker container.

jashapiro: I'm not sure we care about versioned transcript IDs, especially if we generate our own transcriptome file... If we can't get versioned transcripts for unspliced RNAs, then we probably don't want them if we are to do any comparisons? I mean, I don't know how comparable single-nucleus and single-cell will be, but it seems we might want to make an attempt? I have to think about it. 

Either way, we ought to be able to get the mitochrondrial gene list out of the gtf as well, so I think we could avoid AnnotationHub (which is something I think I would prefer).

allyhawkins: Yes totally makes sense. I went ahead and merged them into one and made it so we were generating all the needed files without using AnnotationHub. It should work and give us what we need for now.

allyhawkins:
For now, I am not running this inside the docker container. Additionally, I merged the `build_tx2gene_mito.R` and `make_pre_mrna_fasta.R` into one script, also removing the dependence on AnnotationHub. The expanded fasta and Tx2Gene.tsv is created from an input gtf (currently ensembl-103) and then synced to S3 for use in creating the index, which I will add in a separate PR. 

I grabbed both the spliced and unspliced transcripts to create the expanded files used for snRNA-seq and then subset to only the spliced transcripts for the main Tx2Gene.tsv used in alevin for scRNA-seq.

workflows/rnaseq-ref-index/make_pre_mrna_fasta.R
```diff
@@ -0,0 +1,121 @@
...
+# corresponding gene identifiers
+# first make Tx2Gene for all spliced and unspliced sequences
+full_tx2gene <- eisaR::getTx2Gene(
+  grl, filepath = expanded_tx2gene
+)
+
+# next make Tx2gene for only spliced transcripts 
+
+# get list of all spliced transcript ID's
+spliced_genes = metadata(grl)$corrtx %>%
```
jashapiro, line 105: A bit simpler?
```suggestion
spliced_genes = metadata(grl)$featurelist$spliced
```

workflows/rnaseq-ref-index/make_pre_mrna_fasta.R
```diff
@@ -0,0 +1,121 @@
...
+# subset sequences for only 
+splice_grl = grl[spliced_genes]
+
+# export Tx2Gene for spliced transcripts
+spliced_tx2gene <- eisaR::getTx2Gene(
+  splice_grl, filepath = tx2_gene_out
+)
+
+# reimport gtf to get list of mito genes 
+gtf <- rtracklayer::import(expanded_gtf)
```
jashapiro, line 117: I hate this, but trying to do something more efficient may be hard? Not a big deal since we should run this script infrequently.

workflows/rnaseq-ref-index/make_pre_mrna_fasta.R
```diff
@@ -0,0 +1,121 @@
...
+# write the associated annotations to gtf 
+eisaR::exportToGtf(
+  grl, 
+  filepath = expanded_gtf
+)
+
+# create text file mapping transcript and intron identifiers to 
+# corresponding gene identifiers
+# first make Tx2Gene for all spliced and unspliced sequences
+full_tx2gene <- eisaR::getTx2Gene(
```
jashapiro, line 98: So I am thinking about this file and wondering if it is what we are really going to want, or if we need to modify it before use (and this may affect the gtf too?). This is a question about downstream processing, really. Do we want the genes in the tx2gene to be separated by spliced and unspliced, or only include the spliced (original) ids? I kind of think we do want the original only, as otherwise we would run into some problems where reads would be equally mapped to spliced & unspliced, then we have to combine them later. 

The simplest case is to think about a gene with one transcript (maybe one with no introns). After this procedure, we will have 2 "transcripts": `ENST000` and `ENST000-U` and a bunch of reads with counts that are getting split between them (we better not use any setting that throws out multimapped reads!). Those counts will then be split between gene `ENSG000` and `ENSG000-U`, but they should all be assigned to `ENSG000`, unless we are trying to keep them all separate. If we were doing RNA-velocity, we would want to keep track, of course, but we aren't at least not at this point, so does it make sense to change all the gene ids to drop the `-U`?

Alternatively, we might want to go back to the version with introns separately and then join intron counts in later analysis? How did you handle this in the past? How does CellRanger handle it?

jashapiro (commented):
This looks good for doing what it does, but I have had a bit of a rethink about how we should be handling all of this, most of that is encompassed in my comment below about gene names.

I guess one question is whether we want to preserve the ability to do RNA-velocity analysis in the future, or just collapse down to the gene level.  This is probably a question that requires some downstream analysis/testing, so I might suggest that we make this script make more versions of the relevant files. Rather than just one `expanded` set, perhaps we should do a `spliced_unspliced` and a `spliced_intron` set? Or just make three sets, which can be combined as needed when building indexes. 

I think either way, I would probably make the gene names for unspliced transcripts the same as the spliced gene names, but I could be convinced otherwise.

allyhawkins:
Okay so I'm going to address some of the bigger comments as a whole before going through and making the edits and addressing the smaller comments, because I think some decisions need to be made before modifying the script further. @jashapiro I think you are correct that if we have a spliced and unspliced version of the same transcript ID, we might see some reads getting split between them. I wasn't quite sure thinking about it if we would want spliced + unspliced together or to split it so that we had them in separate fasta's like you suggested, or to do introns rather than unspliced). Looking at the way that cellranger describes their use of a "pre mRNA reference" it looks like they separate the intron reads vs. the exon reads. 

> The single-nuclei RNA-seq assay captures unspliced pre-mRNA as well as mature mRNA. However, after alignment, cellranger count only counts reads aligned to exons. Since the pre-mRNA will generate intronic reads, it may be useful to count these reads as well. Previously, it was recommended to create a custom “pre-mRNA” reference package, listing each gene transcript locus as an exon, in order to count intronic reads. In Cell Ranger 5.0, there is a new include-introns option for counting intronic reads that should be used instead, and the usage of pre-mRNA references is deprecated.

To me this means we would want to use the previous method of getting spliced and intronic reads and then exporting into a fasta. I could then export both the spliced + introns fasta (as one file) and spliced only fasta (i.e. main txome.fa) and corresponding gtfs. 

I believe in that case, unless we are keeping track for RNA velocity, we would only want one Tx2Gene mapping containing the transcript id's for the spliced mRNA mapped to the ensembl gene id's? We could also generate a separate file with the same transcript ID's but with the intron specification added solely to be used if we are to incorporate RNA velocity down the line. 

To address your point about tximeta - in the [alevin tutorial](https://combine-lab.github.io/alevin-tutorial/2020/alevin-velocity/) it appears that by creation of our own annotated gtf from the custom transcriptome rather than the ensembl release might make things easier on us and allow us to use with tximeta. 

> To enable reading the estimated abundances with tximeta, automatically recognizing the underlying transcriptome, we write the expanded annotation to a GTF file. This will later be used to create a linked transcriptome for tximeta.

jashapiro:
> Okay so I'm going to address some of the bigger comments as a whole before going through and making the edits and addressing the smaller comments, because I think some decisions need to be made before modifying the script further.

Agree 100%! (Having the framework of code is good for discussion, but don't modify it more until we know what we want.)

I'm still confused by this:

> In Cell Ranger 5.0, there is a new include-introns option for counting intronic reads that should be used instead, and the usage of pre-mRNA references is deprecated.

Specifically, what is output in this case? Just the total counts by gene, or by transcript? 

My tximeta question was more of a side note, but I think the information there is informative! I am a bit confused though by the tutorial, since it starts with defining introns, but then calls the result of `tximeta::splitSE()` `"spliced"` and `"unspliced"` (oh, I just found where the rename it). It makes sense that intron-only is probably a better estimate of unspliced abundance than if the full unspliced transcript were used (since reads from spliced transcripts would all be split between the two). 

In any case, after `tximeta::splitSE()`, we would have a `"spliced"` and `"unspliced"` counts matrix: Am I correct in thinking we could just sum those to get a "total" matrix for gene counts? Would that be "equivalent" to what CR is doing?

(Also interesting to me is that the `splitSE()` seems to summarize to gene level? They don't explicitly mention that, but it seems to use the gene correspondence table, not transcript correspondence).

allyhawkins:
> Specifically, what is output in this case? Just the total counts by gene, or by transcript?

So it is my interpretation that we would still only get the total counts by gene in which reads mapped to introns are also included in the count. Theoretically it looks like to take a similar approach in alevin and kallisto we would want to add *only introns* (but not the full unspliced transcript) to our transcriptome fasta containing spliced transcripts. You then would use tximeta like you would with single cell RNA seq to obtain a counts x gene matrix without splitting by spliced or unspliced (unless you are doing velocity analysis). 

This is the way I am envisioning it (and also anticipate this is how cellranger works): Let's say you have 10 reads that correspond to gene X, 5 of those are from exonic regions of the gene and 5 are from intronic regions of the gene. When you map to a joint txome + introns you would get a total of 10 reads that map to that gene (regardless of where in the gene they are). So when you import it into tximeta, your matrix would still have 10 reads corresponding to that gene. Now if you use `tximeta::splitSE()` and the corresponding Tx2Gene map that contains the identifiers of spliced vs. intron reads then you can split that matrix up to calculate RNA velocity. But in our case we wouldn't need to do that and would skip that step alltogether. 

I could very much be interpreting how cellranger counts exons vs. introns all wrong, but based on their diagrams for alignment it appears that without `--include-introns` they throw out any read that aligns to an intron in the final count and with that flag they are now including it.

jashapiro:
> This is the way I am envisioning it (and also anticipate this is how cellranger works): Let's say you have 10 reads that correspond to gene X, 5 of those are from exonic regions of the gene and 5 are from intronic regions of the gene. When you map to a joint txome + introns you would get a total of 10 reads that map to that gene (regardless of where in the gene they are). So when you import it into tximeta, your matrix would still have 10 reads corresponding to that gene. Now if you use `tximeta::splitSE()` and the corresponding Tx2Gene map that contains the identifiers of spliced vs. intron reads then you can split that matrix up to calculate RNA velocity.

Oh, that was not my interpretation. I was expecting that `tximeta` import would result in a SingeCellExperiment matrix with two rows for each gene: one  `ENSG000` and one `ENSG000-I`; The `splitSE` turns that into two congruent matrices of half the size, labelled with the `ENSG000` for each gene. As far as `alevin` was concerned, those were two different genes, so it quantifies them separately, and we would have to collapse them ourselves (but splitSE makes this simpler). 

With CellRanger, I think it is probably just doing the gene count? I haven't looked at that manual: I am assuming you are quoting from the single-nucleus instructions; are there different instructions for RNA velocity when they would want to keep the intron counts separate?

allyhawkins:
> Oh, that was not my interpretation. I was expecting that `tximeta` import would result in a SingeCellExperiment matrix with two rows for each gene: one `ENSG000` and one `ENSG000-I`; The `splitSE` turns that into two congruent matrices of half the size, labelled with the `ENSG000` for each gene. As far as `alevin` was concerned, those were two different genes, so it quantifies them separately, and we would have to collapse them ourselves (but splitSE makes this simpler).

Okay so it looks like according to the way that alevin is using `tximeta` and `splitSE` there are 3 assays in the final summarized object. a) a counts assay, which theoretically is going to contain the gene x count matrix separating out the counts by intron vs. exon, b) a spliced only counts matrix, and c) an unspliced (i.e. intron reads) only count matrix. So I think what you are saying and how our ultimate gene x counts matrix can be computed is by doing what you said earlier about splitting the original counts assay into spliced and unspliced and then summing the counts that correspond to the same gene. The question is can we also do this for kallisto and based on my naive understanding of kallisto I think we could? 

All that being said, I think the approach here is to follow a similar approach to the alevin tutorial and produce the following reference files: 

*To be used for snRNA-seq:*
- spliced + introns fasta 
- spliced + introns gtf
- spliced + introns Tx2Gene mapping

*To be used for scRNA-seq:*
- spliced only fasta (i.e. txome.fa)
- spliced only gtf
- spliced only Tx2Gene

*For QC:*
- mitochondrial gene list

jashapiro:
> The question is can we also do this for kallisto and based on my naive understanding of kallisto I think we could?

I think so too, but we will probably have to do some finagling to parse things out. (Or maybe we can use `BUSpaRse` to smooth things a bit). 

I think your plan sounds good, I would just add that we will want to output the `metadata(grl)$corrgene` table with the snRNA-seq annotations to ease later parsing (I'm sure we could generate it on the fly, but that seems unnecessary).

A couple other comments as you finalize this:
- It is probably worth compressing the gtf file & fasta files. It looks like the main output functions won't do that, so we should add that to the wrapper function before syncing to s3.
- Do we want to put the fasta files in the `fasta` directory? I can't say why I kept them separate from annotations before, but we should be consistent somehow...
- Does the spliced gene list have every gene, including noncoding genes? I ask because I assume that those are less likely to have introns, and I just don't know what the `eisaR` package is doing there. I hope that it keeps everything, and single exon genes just have a corresponding "intron" gene with no transcripts, but I fear that they might be dropped. We should do a quick sanity check comparing the gtf (and fasta?) generated by eisaR for spliced only transcripts with the txome gtf from ensembl.

allyhawkins:
The scripts have now been updated to include the following outputs. 

*To be used for snRNA-seq:*

- spliced + introns fasta
- spliced + introns gtf
- spliced + introns Tx2Gene mapping
- *spliced mrna + introns metadata table*

*To be used for scRNA-seq:*

- spliced only fasta (i.e. txome.fa)
- spliced only gtf
- spliced only Tx2Gene

*For QC:* 

- mitochondrial gene list

I have also modified them to include compressing the fasta and gtf files. The fasta files can be compressed as they are exported using `Biostrings::writeXStringSet`, but the gtf files will need to be compressed in the `get-refs.sh` wrapper. 

That being said, I have also added in a notebook in `analysis` that compares the fasta and gtf files that we get when directly fetching from ensembl vs. grabbing just the sequences corresponding to spliced transcripts using `eisaR::getFeatureRanges`.  The analysis is fairly surface level and mainly looks at the intersection of transcript ID's in both the fasta and gtf files, but it is clear that we are missing some transcripts that correspond to both ncRNA and protein coding transcripts. In terms of trying to make sure we are not veering too far of course (from the public standard and in terms of time spent), it seems like we should be sticking with the indexes we have originally made for scRNA-seq and removing those parts of the script, but am definitely open to other suggestions if people have them.

jashapiro (commented):
I first was looking at the missing transcripts, and the first few are very small! Is that a universal feature? If so, we might be fine to lose them.  

From your nb:
```
>head(lost_tx)


DNAStringSet object of length 6:
    width seq                                                                    names               
[1]    12 GGGACAGGGGGC                                                           ENST00000631435
[2]    11 CTAACTGGGGA                                                            ENST00000632524
[3]    20 GGTATAGTGGGAGCTACTAC                                                   ENST00000633009
[4]    18 GGGTATAGCAGCGGCTAC                                                     ENST00000634070
[5]    20 GTAGAGATGGCTACAATTAC                                                   ENST00000632963
[6]    19 TGACTACGGTGGTAACTCC                                                    ENST00000633030`
```

I was confused by `missing_gtf`, because there were transcripts in the listed files that were in both, but after a lot of head scratching,  I realized that there is an error in the code, which explains that, which I have noted in a line comment, but appears here: https://github.com/AlexsLemonade/alsf-scpca/blob/f863e68af9ad6a41ac218c2f76f43edddbbd572f/analysis/reference-file-comparisons/compare_fasta_gtf.Rmd#L148-L149 


I also left a couple of other line-level comments, mostly about file naming. On that front, do we want to rename the original download fasta files to add the ensembl version number? 

 There are also a few little suggestions left from my last review.

workflows/rnaseq-ref-index/make_pre_mrna_fasta.R
```diff
@@ -0,0 +1,171 @@
...
+
+## files with spliced + intron regions
+spliced_intron_fasta_file <- paste0(file_base,".", "spliced_intron.txome.fa.gz")
+spliced_intron_gtf_file <- paste0(file_base, ".", "spliced_intron.txome.gtf")
+spliced_intron_tx2gene_file <- paste0(file_base, ".", 
+                                      "spliced_intron.tx2gene.tsv")
+spliced_intron_metadata_file <- paste0(file_base, ".", 
+                                       "spliced_intron.metadata.tsv")
+
+mito_file <- paste0(file_base, ".", "mitogenes.txt")
```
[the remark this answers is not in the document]
allyhawkins: No reason not to at all. Just me completely missing that and making it harder on myself, whoops!

workflows/rnaseq-ref-index/make_pre_mrna_fasta.R
```diff
@@ -0,0 +1,169 @@
...
+)
+
+# create text file mapping transcript and intron identifiers to 
+# corresponding gene identifiers
+# first make Tx2Gene for all spliced and intron sequences
+full_tx2gene <- eisaR::getTx2Gene(
+  grl, filepath = spliced_intron_tx2gene
+)
+
+# next make Tx2gene for only spliced transcripts
```
jashapiro, line 133: This comment somehow got left after its useful time?
```suggestion
```

workflows/rnaseq-ref-index/get-refs.sh
```diff
@@ -17,6 +19,16 @@ aws s3 sync fasta $s3_base/fasta
...
 
+# running without docker for now due to docker error
+# $rdocker Rscript make_pre_mrna_fasta.R
+
+Rscript make_pre_mrna_fasta.R
+gzip annotation/*.gtf
+
+# the sync with fasta would need to happen here if we are to remove the
+# previous step of combining cdna & ncrna fasta files
+#aws s3 sync fasta $s3_base/fasta
```
jashapiro, line 31: I think we do want to sync this folder too now that the fasta files go here?
```suggestion
aws s3 sync fasta $s3_base/fasta
```

jashapiro (commented):
I think this looks just about ready to go in. We can evaluate the annotations we get separately, as you have started to do in #73, but the script looks mostly done. 

My remaining suggestions are the ones below here, and the prior comment about alphabetizing the packages in the Dockerfile. I don't particularly care if you use stringr::word from my other lingering suggestion, but the other things should get done before merging.

allyhawkins:
> I don't particularly care if you use stringr::word from my other lingering suggestion, but the other things should get done before merging.

Sorry I missed that in the previous review. It's incorporated now along with the other changes. All the files should be synced and on AWS now as well.

jashapiro (approved):
LGTM!

PR #70, opened 2021-04-07
Update cellranger to allow flexibility for all 10x technology

allyhawkins:
Similar to #52, the cellranger workflow also needed to be updated to run with any of the 10x technologies (v2, v3, or v3.1). This is now included in the workflow.

Additionally, there was an error in the getCRsamples function that threw an indexing error if the folders where the fastq files lived also contained index files. To go around this, the regular expression has been altered to find the sample names from both the R1/R2 fastqs or index fastqs.

### DIFF
--- a/workflows/cellranger-quant/run-cellranger.nf
+++ b/workflows/cellranger-quant/run-cellranger.nf
@@ -13,7 +13,7 @@ params.run_ids = "SCPCR000001,SCPCR000002"
 params.outdir = 's3://nextflow-ccdl-results/scpca/cellranger-quant'
 
 // technology options
-technology = ["10Xv2", "10Xv3", "10Xv3.1"]
+tech_list = ["10Xv2", "10Xv3", "10Xv3.1"]
 
 // build full paths
 params.index_path = "${params.ref_dir}/${params.index_dir}/${params.index_name}"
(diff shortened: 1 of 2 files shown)


workflows/cellranger-quant/run-cellranger.nf
```diff
@@ -12,6 +12,9 @@ params.run_metafile = 's3://ccdl-scpca-data/sample_info/scpca-library-metadata.t
 params.run_ids = "SCPCR000001,SCPCR000002"
 params.outdir = 's3://nextflow-ccdl-results/scpca/cellranger-quant'
 
+// technology options
+technology = ["10Xv2", "10Xv3", "10Xv3.1"]
```
jashapiro, line 16: Minor thing that I try to do, but you might not care about, is that if there is a list/vector I try to give it a plural name or some kind of suffix to indicate the data structure.
```suggestion
tech_list = ["10Xv2", "10Xv3", "10Xv3.1"]
```

workflows/cellranger-quant/run-cellranger.nf
```diff
@@ -60,7 +63,7 @@ workflow{
   run_all = run_ids[0] == "All"
   ch_reads = Channel.fromPath(params.run_metafile)
     .splitCsv(header: true, sep: '\t')
-    .filter{it.technology == "10Xv3"} // only 10X data
+    .filter{it.technology in technology} // only 10X data
```
jashapiro, line 66: If you like my previous change, I would make this one too.
```suggestion
    .filter{it.technology in tech_list} // only 10X data
```

jashapiro (approved):
Looks good. Just one suggestion on variable names for clarity.

PR #71, opened 2021-04-07
Add 10X support  to alevin and alevin-fry and --sketch option

### LINKED ISSUE #52: Add flexibility for 10Xv2 data

### LINKED ISSUE #64: Add --sketch option to alevin-fry workflow
One of the changes that came with `alevin-fry` was a new mapping option in `salmon alevin` that performs pseudomapping that is much more similar to `kallisto`. This does make it much faster, so it is an option we might well want to explore in benchmarking (#63). I am here proposing that we make this a workflow option using nextflow `params` so that we can easily switch between the two for benchmarking.

Note that we will also probably want to add something to the output directory (for `publishDir` flags) or the work directory currently defined as `run_dir` so that we do not accidentally clobber previous results.

jashapiro:
This PR does two things: 

1. Adds the flexibility to alevin and alevin-fry workflows to allow them to handle 10Xv2 and v3.1 data. This also fixes an error we had in invocation of alevin, where we should have been using the `--chromiumV3` flag. I don't know what difference that was making, but presumably we had a slightly incorrect definition of UMI and/or cell barcode locations. Because the other caller workflows already have the v2 flexibility, this PR closes #52

2. Adds a `--sketch` option to the alevin-fry workflow. This is invoked with the `--sketch` flag at workflow invocation. One question is whether we want to make this the default for alevin-fry, in which case we might change the flag to be something like `--traditional`. I think this is fine for now, but depending on results of benchmarking, we might switch it in the future. Another option is to make it a nextflow `params` option with a value (`--mode sketch` vs. `--mode selaln` or something like that), but I thought a simple flag would work here. Closes #64


allyhawkins (approved):
LGTM!

PR #73, opened 2021-04-14
Comparison of Reference File Generation

allyhawkins:
I'm separating out the analysis started in #68 as a separate PR, to just bring attention to making a final decision on whether or not we should alter how we are generating our reference files for creating indexs using in the scpca workflows. Currently we are using downloading the cdna.fa and ncrna.fa files from ensembl directly and then concatenating them to create a transcriptome.fa file to be used to generate indexs for Alevin and Kallisto for scRNA-seq pre-processing.

In generating the reference files for snRNA-seq for Alevin and Kallisto, we now need to grab the primary_assembly.fa and gtf, subset the spliced mRNA and flanking introns using `eisaR::getFeatureRanges`, and then output a corresponding fasta, gtf, and transcript to gene mapping. The question being posed is can we maintain consistency between our references for both snRNA-seq and scRNA-seq by now doing the same thing for the main transcriptome.fa rather than concatenating the files. 

To determine this, I have compared the fasta's and gtf files that come directly from ensembl vs. from using `eisaR::getFeatureRanges` on the primary_assembly.fa and gtf files. In my analysis, I am finding that these files are not the same, and there are transcripts missing that correspond to protein coding regions. Due to these discrepancies, I think it is in our best interest to use our original reference files and only apply the `eisaR::getFeatureRanges` method to get reference files needed for snRNA-seq. It appears that is the best way to do that right now. As long as there are no objections to this, then the reference file scripts in #68 will be used to generate only the files needed for snRNA-seq.


analysis/reference-file-comparisons/compare_fasta_gtf.Rmd
```diff
@@ -127,58 +155,53 @@ genome_gtf <- genome_gtf[genome_gtf$transcript_id %in%
...
-                          genome_gtf$transcript_id]
+shared_gtf <- genome_gtf[genome_gtf %in% txome_gtf]
 
 # get the features only found in the eisaR txome
-txome_only_gtf <- txome_gtf[!txome_gtf$transcript_id %in% 
-                              genome_gtf$transcript_id]
+txome_only_gtf <- txome_gtf[!txome_gtf %in% genome_gtf]
+
+# get the features only found in the ensembl gtf
+genome_only_gtf <- genome_gtf[!genome_gtf %in% txome_gtf]
```
jashapiro, line 164: Why did you switch here to the full entries, rather than `transcript_id`? Does this make a difference in the counts?

allyhawkins: I was noticing that I was missing some of the overlap if I just did the transcript ID's and that it did make a difference in the counts, so it seemed more accurate to look at the entire entry rather than just the transcript ID.

jashapiro: I guess I am not clear on what the comparison actually is then; comparing objects can have unexpected behavior, and I wonder if this might be inflating counts. 

A secondary question, which I meant to ask you but hadn't yet, is how many genes are actually affected? Are we losing individual transcripts from genes, or entire genes?

allyhawkins: So it looks like looking specifically at the gene level, there are no genes lost between the ensembl version and eisaR version. However, there are some genes that are gained in the eisaR version, not sure that gaining uncategorized genes is all that relevant though. The transcript level seems to be where the discrepancies are, so the question is do we care about that or not?

jashapiro: I think the ultimate test is going to be to compare mapping to an index generated from the eisaR spliced fasta to the one generated from the ensembl reference fasta. If they are identical, then we don't care. If they differ...

allyhawkins: Very good point. I'm going to add in the workflows for generating the pre mRNA index files and then we can go from there in terms of comparing the eisaR spliced fasta to the ensembl one.

allyhawkins:
In an effort to not keep duplicate analysis, I removed the first version of the notebook so there is just the one notebook containing reference file comparisons to merge so we can close #54.

jashapiro (approved):
Since you already reviewed this when I wrote it, it seems silly for me to make any new comments!

PR #75, opened 2021-04-16
Modify index building workflows to include pre mRNA index 

allyhawkins:
In this PR, I am altering the current `build-index.nf` workflow to build salmon and kallisto indexes using both the `spliced.txome.fa` and the `spliced_intron.txome.fa` that were generated in #68. Additionally, I updated all of the index's to now use the ensembl-103 build and the salmon index to use the container with the most recent salmon release (also used in the `alevin-quant/run-alevin.nf` workflow). 

One thing to note is that generating the index using the `spliced_intron.txome.fa` takes considerably more memory, and that has also now been adjusted so the workflow runs successfully.

### DIFF
--- a/workflows/rnaseq-ref-index/build-index.nf
+++ b/workflows/rnaseq-ref-index/build-index.nf
@@ -3,7 +3,6 @@ nextflow.enable.dsl=2
 
 // basic parameters
 params.ref_dir = 's3://nextflow-ccdl-data/reference/homo_sapiens/ensembl-103'
-params.cdna = 'fasta/Homo_sapiens.GRCh38.cdna.all.fa.gz'
 params.spliced_txome = 'fasta/Homo_sapiens.GRCh38.103.spliced.txome.fa.gz'
 params.spliced_intron_txome = 'fasta/Homo_sapiens.GRCh38.103.spliced_intron.txome.fa.gz'
 params.ensembl_txome = 'fasta/Homo_sapiens.GRCh38.txome.fa.gz'
@@ -39,7 +38,6 @@ process salmon_index_no_sa{
       -i ${index_base}_k${kmer} \
       -k ${kmer} \
       -p ${task.cpus} \
-      --gencode
     """
 }
 
@@ -68,7 +66,6 @@ process salmon_index_full_sa{
       -i ${index_base}_k${kmer}_full_sa \
       -k ${kmer} \
       -p ${task.cpus} \
-      --gencode
     """
 }
 
@@ -95,8 +92,7 @@ process kallisto_index{
 workflow {
   // channel of the reference files and labels
   ch_ref = Channel
-    .fromList([["cdna", params.ref_dir + "/" + params.cdna, params.kmer],
-               ["ensembl_txome", params.ref_dir + "/" + params.ensembl_txome, params.kmer],
+    .fromList([["ensembl_txome", params.ref_dir + "/" + params.ensembl_txome, params.kmer],
                ["spliced_txome", params.ref_dir + "/" + params.spliced_txome, params.kmer],
                ["spliced_intron_txome", params.ref_dir + "/" + params.spliced_intron_txome, params.kmer]])
 
--- a/workflows/rnaseq-ref-index/build-index.nf
+++ b/workflows/rnaseq-ref-index/build-index.nf
@@ -2,9 +2,11 @@
 nextflow.enable.dsl=2
 
 // basic parameters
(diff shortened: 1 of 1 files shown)


workflows/rnaseq-ref-index/build-index.nf
```diff
@@ -32,12 +37,14 @@ process salmon_index_no_sa{
     salmon index \
       -t ${reference} \
       -i ${index_base}_k${kmer} \
-      -k ${kmer}
+      -k ${kmer} \
+      -p ${task.cpus} \
+      --gencode
```
jashapiro, line 42: Do we need this flag? I don't think any of our fasta files use gencode formatting.

workflows/rnaseq-ref-index/build-index.nf
```diff
@@ -2,9 +2,11 @@
 nextflow.enable.dsl=2
 
 // basic parameters
-params.ref_dir = 's3://nextflow-ccdl-data/reference/homo_sapiens/ensembl-100'
+params.ref_dir = 's3://nextflow-ccdl-data/reference/homo_sapiens/ensembl-103'
 params.cdna = 'fasta/Homo_sapiens.GRCh38.cdna.all.fa.gz'
```
jashapiro, line 6: We should probably remove this; I don't think we are likely to use a cdna-only index given earlier testing. We almost certainly want the ncRNA!

workflows/rnaseq-ref-index/build-index.nf
```diff
@@ -2,9 +2,11 @@
...
-params.txome = 'fasta/Homo_sapiens.GRCh38.txome.fa.gz'
+params.spliced_txome = 'fasta/Homo_sapiens.GRCh38.103.spliced.txome.fa.gz'
+params.spliced_intron_txome = 'fasta/Homo_sapiens.GRCh38.103.spliced_intron.txome.fa.gz'
+params.ensembl_txome = 'fasta/Homo_sapiens.GRCh38.txome.fa.gz'
```
jashapiro, line 9: This is probably fine for now, but if we find we are changing the inputs like this too often, we might want to just make the workflow more flexible, and allow specifying a list of transcriptome files and a parallel list of ids, rather than hard coding the IDs for each kind of transcriptome

workflows/rnaseq-ref-index/build-index.nf
```diff
@@ -60,14 +67,18 @@ process salmon_index_full_sa{
...
 }
 
 process kallisto_index{
   container 'quay.io/biocontainers/kallisto:0.46.2--h4f7b962_1'
   publishDir "${params.ref_dir}/kallisto_index", mode: 'copy'
-  memory 32.GB
+  memory { 120.GB * task.attempt}
+  cpus 8
+  errorStrategy { task.exitStatus in 137..140 ? 'retry' : 'terminate' }
+  maxRetries 1
```
jashapiro, line 81: Point of curiousity... did you ever see it retry?

workflows/rnaseq-ref-index/build-index.nf
```diff
@@ -60,14 +67,18 @@ process salmon_index_full_sa{
...
+      --gencode
     """
 }
 
 process kallisto_index{
   container 'quay.io/biocontainers/kallisto:0.46.2--h4f7b962_1'
   publishDir "${params.ref_dir}/kallisto_index", mode: 'copy'
-  memory 32.GB
+  memory { 120.GB * task.attempt}
+  cpus 8
```
jashapiro, line 79: Just to note that these are probably redundant here: once you get to a 120GB memory machine, it already has at least 16 cores, based on https://aws.amazon.com/ec2/instance-types/ (I think it will choose an R5a machine when it needs a lot of memory.)

jashapiro (approved):
This looks good to me. I'm surprised but not shocked by the amount of memory needed by kallisto. I have some thoughts on future modifications for more flexibility, but this workflow is not meant to be run very often, so we probably don't need to do those for now.

workflows/rnaseq-ref-index/build-index.nf
allyhawkins: Yes definitely agreed. Maybe if we start changing things around more we can alter it to just have a fasta input and corresponding name like you said. Perhaps something to change if we try the toplevel.fa...

workflows/rnaseq-ref-index/build-index.nf
allyhawkins: I don't see anything in the tower output that indicates it retried... But it sure did fail with the 137 error quite a few times.

PR #76, opened 2021-04-16
Add --include-introns for cellranger workflow

### LINKED ISSUE #72: Add --include-introns as option to cellranger workflow
Currently the workflow for cellranger would not be compatible for use with the single-nuclei RNA-seq samples and would throw out any reads aligned to introns. To ensure that we can use cellranger for both snRNA-seq and scRNA-seq, we need to add in the `include-introns` option as a parameter to the `workflows/cellranger-quant/run-cellranger.nf` workflow. 

See this information from cellranger on using the `include-introns` flag for [reference](https://support.10xgenomics.com/single-cell-gene-expression/software/pipelines/latest/advanced/references):

> The single-nuclei RNA-seq assay captures unspliced pre-mRNA as well as mature mRNA. However, after alignment, cellranger count only counts reads aligned to exons. Since the pre-mRNA will generate intronic reads, it may be useful to count these reads as well. Previously, it was recommended to create a custom “pre-mRNA” reference package, listing each gene transcript locus as an exon, in order to count intronic reads. In Cell Ranger 5.0, there is a new include-introns option for counting intronic reads that should be used instead, and the usage of pre-mRNA references is deprecated.

allyhawkins:
Closes #72. I added a parameter to the run-cellranger.nf workflow to allow for flexibility to use cellranger with snRNA-seq samples. To use the --include-introns flag you would implement it at the time of running the workflow by executing: 

`nextflow run cellranger-quant/run-cellranger.nf --include_introns:"--include-introns"`

I don't think this is a permanent fix for this. Before we run all of our samples I think a better solution might be to grab the `seq-unit` from the `scpca-library-metadata.tsv` file and based on if the `seq-unit == nucleus`, then we would use the flag for `--include-introns`. 

However, since we are still in the benchmarking phase, I thought it might not be the best approach to give up complete control. It could be useful to compare some of the snRNA-seq runs using cellranger with and without this flag rather than include it as a default immediately.

### DIFF
--- a/workflows/cellranger-quant/run-cellranger.nf
+++ b/workflows/cellranger-quant/run-cellranger.nf
@@ -11,11 +11,10 @@ params.run_metafile = 's3://ccdl-scpca-data/sample_info/scpca-library-metadata.t
 // or "All" to process all samples in the metadata file
 params.run_ids = "SCPCR000001,SCPCR000002"
 params.outdir = 's3://nextflow-ccdl-results/scpca/cellranger-quant'
+params.include_introns = false // use --include-introns to align to pre mRNA index
 
 // technology options
 tech_list = ["10Xv2", "10Xv3", "10Xv3.1"]
-
-params.include_introns = false 
 
 // build full paths
 params.index_path = "${params.ref_dir}/${params.index_dir}/${params.index_name}"
--- a/workflows/cellranger-quant/run-cellranger.nf
+++ b/workflows/cellranger-quant/run-cellranger.nf
@@ -22,15 +22,17 @@ params.index_path = "${params.ref_dir}/${params.index_dir}/${params.index_name}"
 process cellranger{
   container '589864003899.dkr.ecr.us-east-1.amazonaws.com/scpca-cellranger:6.0.0'
   publishDir "${params.outdir}", mode: 'copy'
+  tag "${id}-${index}" // add tag for tracking sample names in trace file  
   label 'cpus_8'
   label 'bigdisk'
   input:
     tuple val(id), val(samples), path(fastq_dir)
     path index
+    val include_introns
   output:
     path output_id
   script:
-    output_id = "${id}-${index}-${params.include_introns ? '--include-introns' : 'pre_mRNA'}"
+    output_id = "${id}-${index}-${include_introns ? 'pre_mRNA' : ''}"
     """
     cellranger count \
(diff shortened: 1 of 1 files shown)


workflows/cellranger-quant/run-cellranger.nf
```diff
@@ -29,15 +30,17 @@ process cellranger{
   output:
     path output_id
   script:
-    output_id = "${id}-${index}"
+    output_id = "${id}-${index}-${params.include_introns ? '--include-introns' : 'pre_mRNA'}"
```
jashapiro, line 33: I assume you want to distinguish here with something like `pre_mRNA` vs `mRNA`? (Not sure what the appropriate alternative is... Maybe just `pre_mRNA` vs nothing? In any case, I think this needs to be updated to reflect the chosen option accurately. Maybe: 
```suggestion
    output_id = "${id}-${index}${params.include_introns ? '-pre_mRNA' : ''}"
```

Or if we want to name both versions explicitly:
```suggestion
    output_id = "${id}-${index}-${params.include_introns ? 'pre_mRNA' : 'mRNA'}"
```

jashapiro (commented):
Looks good, with one suggestion based on the current structure of the PR. 

But I have a question that is probably more for later: Should we automatically switch modes based on the sample? That is, if the `seq_unit` is `nucleus`, we use `--include-introns` but if it is `cell` we do not? I am assuming that for testing for now we want to be able to switch manually, but for later, we may want to include that as an option. If we do want that flexibility later, it may make sense to use include the option as a `input` to the process (and pass it in as part of the workflow), rather than using the `params` value directly.

jashapiro:
> Should we automatically switch modes based on the sample? That is, if the `seq_unit` is `nucleus`, we use `--include-introns` but if it is `cell` we do not?

I just realized that you addressed this in your PR submission, which I obviously didn't read again this morning. I blame Monday. Apologies. 

I still think it makes sense to set the option up as an input to the process to facilitate later changes, but I'm really sorry for just repeating something you already addressed.

allyhawkins:
No worries at all. I do agree we will need to do a final update before we run everything to automatically use the --include-introns for snRNA-seq samples, assuming we that's what we end up deciding on. I moved it to be an input to the process to get us started.

jashapiro (approved):
LGTM!

PR #77, opened 2021-04-19
Fixed a minor error in where the mito gene list is output in the reference fasta script 

allyhawkins:
In working through some of the benchmarking analysis, I realized that the mitogenes.txt file that is made in the new `make_pre_mrna_fasta.R` script wasn't output to the annotation folder like it should have been so was never synced to S3. I made the minor fix and then re-synced to S3 so it should be there.


jashapiro (approved):
LGTM!

PR #78, opened 2021-04-21
Minor fixes to naming of flag for --include-introns in cellranger workflow

allyhawkins:
I found some minor errors in the naming of the `--include-introns` flag that was causing errors with writing output files. Additionally, I have altered the workflow slightly to make it easier to alter when we want to automate it to use the `--include-introns` flag for all snRNA-seq samples. 

To now implement the `--include-introns` flag in cellranger count, you should be able to use the flag `--premrna` at the command line.


jashapiro (approved):
LGTM!

Issue #63, opened 2021-03-31
Benchmarking Plan for ScPCA


allyhawkins:
Closing #9 and #59, the previous issues on benchmarking preprocessing workflows, and opening a new issue to start discussing a concrete plan on building the tools we need for benchmarking. 

In thinking about the benchmarking, it appears that the main goal is to identify a pre-processing workflow that is going to be computationally efficient and result in the most accurate alignment and assignment of UMI’s to cells. I think there are a variety of ways that we could test this to different extremes, but it looks like there are some basic QC metrics that we would want to output for each sample: 

- Knee plot (distribution of UMIs/cell)
- Mitochondrial content of cells 
- UMIs/gene
- genes/cell 

It appears that one nice way of getting many of these key metrics is using addPerCellQC in the scater Bioconductor package. Additionally we could use miQC which is designed to run with this function to make informed decisions on how to filter poor quality cells based on both mitochondrial content and number of UMI’s per cell. This would be a good workflow to explore and then create uniform plots across the samples for easy comparisons. Some ways to make direct comparisons can be to look at the following: 

- Correlation of average gene expression/cell across different tools in the same sample 
- Plot # UMI on x-axis and #gene on y-axis for all cells and show cutoff of miQC in each of the pre-processing workflows (i.e. for each workflow how many cells are we getting that are considered “poor”quality? - see [figure 2c](https://www.biorxiv.org/content/10.1101/673285v2.full))  
- Processing time and Memory usage vs. Number of reads 

In addition to making these comparisons across tools, we are also interested in finding a tool that will be consistent regardless of confounding variables that are present in our dataset, i.e. library preparation, tissue type, or choice of technology. In order to address these questions, we should make sure we have some samples from each of these groups. 

After going through the samples that are currently listed in the `ScPCA-admin/sample_info/scpca-sample-metadata.tsv` and corresponding library information in `scpca-library-metadata.tsv`, we have 131 runs that are from single-cell vs. 16 runs from single-nucleus RNA-seq. All 16 snRNA-seq runs are from neuroblastoma, so it appears that to address any changes between single cell vs. single nucleus we should compare NB single-cell vs. NB single-nucleus samples. 

[...]

PR #80, opened 2021-04-21
Preliminary benchmarking of Memory/Time footprint for pre-processing tools

allyhawkins:
In reference to the discussion in #63, here is an initial start to benchmarking specifically focused on comparing computation time and memory usage across the different pre-processing tools. To access the metrics from each nextflow run, I added the `--with-trace` option to `nextflow run` which outputs a `trace.txt` file containing runtime, % CPU, and memory usage. 

Included in this notebook, I have run 4 samples (SCPCR000003, SCPCR000006, SCPCR000126, and SCPCR000127) through all 4 tools. All 4 of these are 10Xv3 and from solid tumors. It is pretty apparent that alevin-fry is using much less memory than the others (at or less than 5 GB/ sample), and Kallisto is the fastest. Cellranger still takes a lot of time and uses more memory. Additionally, it looks like cellranger is the only tool that may increase in runtime with increased number of reads/ sample? It is the tool with the largest variation in runtime across samples. 

Next, I will add some snRNA-seq samples for comparison as they represent about 50% of our samples. I will also add another PR to address differences in quantification across the tools when that is finished. For now this seemed sufficient to get our discussion started.


rob-p:
Hi @allyhawkins, I look forward to learning what insights can be drawn from your benchmarks. 
I'm  curious if the speed measurements here are using alevin-fry in `--sketch` mode or with standard selective alignment.  Thanks!

allyhawkins:
> I'm curious if the speed measurements here are using alevin-fry in `--sketch` mode or with standard selective alignment, and with the `cr-like` (more simplistic) resolution, or the much more heavyweight `full` resolution.

Thanks for the insight! Right now, we are testing with and without `--sketch` mode using the `full` resolution mode. We definitely are seeing speed improvements, even with the `full` resolution, in compared to Alevin (and other tools) thus far. Although I will say this is only on a small subset of samples currently. As we dig in a little further, we will look at exploring different strategies like the `cr-like` mode that you suggested and the `--unfiltered-pl` that you mentioned earlier today. Thanks again!

rob-p:
Thanks for the explanations, @allyhawkins!  Btw, I think that v0.2.0 of alevin-fry should eliminate the small dependence you previously saw of the memory usage on the number of reads.  We are able to process the pbmc10k_v3 data using this version with a peak memory usage of ~2.5G (even in unfiltered mode).

analysis/quantifier-comparisons/03-memory_time_comparison.Rmd
```diff
@@ -0,0 +1,287 @@
...
+```
+
+From this, it looks like there is no effect on the number of reads found in 
+the fastq files to the memory that is required for any of the processes, except
+maybe with alevin-fry both with/without --sketch. It appears that the memory
+usage almost doubles from ~2-3 GB to closer to ~5-6 GB. Even though memory
+usage doesn't appear to be that affected, run time does appear to be affected
+by the number of reads in the sample if running on cellranger. As the number
+of reads increase, there is a sharp increase in the run time for cellranger, 
+where as that is not the case for the other tools.
```
jashapiro, line 275: I think there may be an issue of scaling here, where the increases for alevin-fry and kallisto are there, but not as large. I also wonder if something about the machines chosen by AWS might have varied for Alevin/Alevin-fry in the first case. Seems odd that there wouldn't be an effect of the number of reads on run time, but maybe something else about the data is affecting runtimes.

jashapiro (approved):
This looks good to me. I had one comment about the runtime results, more of a thought than anything else.

analysis/quantifier-comparisons/03-memory_time_comparison.Rmd
allyhawkins: Yes, it does look like the number of reads may also affect Alevin-fry, but the scales make it less visually obvious. I also think there might be clearer trends with more samples since there are only 4 samples here.

PR #84, opened 2021-04-27
Adjust default index references for workflows

allyhawkins:
I am updating the default index's for each of the workflows to be the `spliced_txome_k31` index that was generated from the spliced only cDNA fasta as in #68. Currently, in order to switch to use the `spliced_intron_txome_k31` you will need to do so at the command line by doing `nextflow run alevin-quant/run-alevin.nf --index_name spliced_intron_txome_k31`. I have also updated the default `t2gene.tsv` files to use the `Homo_sapiens.GRCh38.103.spliced.tx2gene.tsv` file. 

Additionally, while running the workflows with the snRNA-seq data, I noticed that they required more memory than previously. The pre-mRNA index generated by Kallisto, in particular, is 45 gb and requires ~ 100 gb of memory for the samples that I did test runs on. Otherwise it is unable to load the index and write the output files. I am using 120 gb of memory here to be safe. 

It appears that Alevin also requires slightly more memory than the 28 gb allotted by the cpus_8 label in the configuration file as well, but nowhere near the 100 gb that Kallisto requires. 

I did spend some time trying to assign individual labels based on the input variable so that we could only use the higher memory requirements for Kallisto with snRNA-seq samples only, but it appears that there is no way to use input variables as  label assignments... https://github.com/nextflow-io/nextflow/issues/894. There is one potential work around that I was able to find to maybe include a second configuration file based on an input parameter, but after some back and forth with that, I landed on leaving it as is for now. 

I am leaving this in draft stage right now, as I still need to change the references in alevin-fry and ensure that workflow runs as expected.

### DIFF
--- a/workflows/kallisto-quant/run-kallisto.nf
+++ b/workflows/kallisto-quant/run-kallisto.nf
@@ -26,10 +26,11 @@ params.mito_path = "${params.ref_dir}/${params.annotation_dir}/${params.mitolist
 process kallisto_bus{
   container 'quay.io/biocontainers/kallisto:0.46.2--h4f7b962_1'
   label 'bigdisk'
-  label 'bigmem'
+  memory { sequnit == 'nucleus' ? '120.GB' : '28.GB'}
+  cpus 8
   tag "${id}-${index}"
   input:
-    tuple val(id), val(tech), path(read1), path(read2)
+    tuple val(id), val(tech), path(read1), path(read2), val(sequnit)
     path index
   output:
     path run_dir
@@ -133,7 +134,8 @@ workflow{
     .map{row -> tuple(row.scpca_run_id,
                       row.technology,
                       file("s3://${row.s3_prefix}/*_R1_*.fastq.gz"),
-                      file("s3://${row.s3_prefix}/*_R2_*.fastq.gz")
+                      file("s3://${row.s3_prefix}/*_R2_*.fastq.gz"),
+                      row.seq_unit
                       )}
   barcodes_ch = samples_ch
     .map{row -> file("${params.barcode_dir}/${barcodes[row.technology]}")}
--- a/workflows/kallisto-quant/run-kallisto.nf
+++ b/workflows/kallisto-quant/run-kallisto.nf
@@ -25,7 +25,8 @@ params.mito_path = "${params.ref_dir}/${params.annotation_dir}/${params.mitolist
 
 process kallisto_bus{
   container 'quay.io/biocontainers/kallisto:0.46.2--h4f7b962_1'
-  label 'bigdisk' 
+  label 'bigdisk'
+  label 'bigmem'
   tag "${id}-${index}"
   input:
     tuple val(id), val(tech), path(read1), path(read2)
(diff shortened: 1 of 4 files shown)


workflows/kallisto-quant/run-kallisto.nf
```diff
@@ -23,10 +23,9 @@ params.index_path = "${params.ref_dir}/${params.index_dir}/${params.index_name}"
 params.t2g_path = "${params.ref_dir}/${params.annotation_dir}/${params.t2g}"
 params.mito_path = "${params.ref_dir}/${params.annotation_dir}/${params.mitolist}"
 
-
 process kallisto_bus{
   container 'quay.io/biocontainers/kallisto:0.46.2--h4f7b962_1'
-  label 'cpus_8'
```
jashapiro, line 29: I think we can have multiple labels? https://www.nextflow.io/docs/latest/process.html#label If so, I would prefer that we separate disk and memory/cpu labels. But in this case, since only this module requires more memory, we probably can specify the memory and cpu specifically for this process (the configuration is not shared by other processes).

workflows/nextflow.config
```diff
@@ -22,6 +22,15 @@ profiles{
       }
       withLabel: bigdisk {
         queue = 'nextflow-batch-bigdisk-queue'
+        cpus = 8
+        memory = { 120.GB * task.attempt}
+        errorStrategy = {task.exitStatus in 137..140 ? 'retry' : 'terminate' }
+        maxRetries = 1
```
jashapiro, line 28: This should probably be put in a separate label (though as I said above, if it is only one process, we might not want to use a label/config). I wanted to keep the queue specification separate from cpu & memory, as there may be processes that need a large amount of disk space but don't need the memory.

workflows/kallisto-quant/run-kallisto.nf
allyhawkins: Yes we can, so I could adjust it to have the bigdisk label with memory referenced in the process. Then the config file wouldn't be altered.

jashapiro: I have no problem with adding more labels, if we want to go that route too.... We could even separate cpu number from memory, but we just want to be careful about trying too hard to optimize. I think having more processes with similar resource requests is more efficient in general.

workflows/nextflow.config
```diff
@@ -23,6 +23,12 @@ profiles{
       withLabel: bigdisk {
         queue = 'nextflow-batch-bigdisk-queue'
       }
+      withLabel: bigmem {
```
jashapiro, line 26: While I think this is fine for testing, if we use this for production, we won't be able to adjust depending on the index we are using. If we keep the memory statement in the process (rather than a tag), I think we may find it easier to change on a per-sample basis (different for nuclei vs cells). At this point, I don't think it matters much, and having the label here doesn't hurt anything, so it is fine.

jashapiro (approved):
LGTM!

Just the forward-looking comment about labels vs. per-process definitions for flexibility.

jashapiro (approved):
🚀

PR #85, opened 2021-04-29
Updates to Alevin-fry workflow to include unfiltered permit list

### LINKED ISSUE #83: Add --unfiltered permit list option to alevin-fry workflow
In reference to the below comment about the new release of Alevin-fry, we should incorporate being able to use the -unfiltered parameter. This should eliminate identifying any barcodes outside the possible list of barcodes in the 10X whitelist. 

Additionally, we will need to update the docker container to use the new release of Alevin-fry. 

Hi all,

  I'll just note that we've recently released [v0.2.0](https://github.com/COMBINE-lab/alevin-fry/releases/tag/v0.2.0) of `alevin-fry`, and it's been propagated through bioconda as well.  In addition to general improvements to the runtime profile (vastly improved memory usage at very high thread count), this release also adds initial (beta) support for quantification via an "unfiltered" permit list using the `-u` option to the `generate-permit-list` command; this functionality was previously unsupported.  This flag will allow one to properly pass e.g. the 10x permit list to the `generate-permit-list` so that one can then quantify all cells, rather than just those selected by the knee threshold.  There is also `--min-reads` command that determines the minimum number of reads (not UMIs, as this is pre-quantification) a cell barcode needs to have to be considered present.  The default is 10, but you can set it at 1 if you really want to quantify _everything_.  Let me know if you need any help or support running alevin-fry on the testing data, or if you have any questions about the different resolution options (e.g. https://github.com/AlexsLemonade/alsf-scpca/pull/60#discussion_r611078240).

Best,
Rob

_Originally posted by @rob-p in https://github.com/AlexsLemonade/alsf-scpca/issues/63#issuecomment-824228309_

allyhawkins:
Closes #83. In this PR, I updated Alevin-fry to use the new Alevin-fry 0.2.0, adding an option to use the unfiltered permit list of barcodes from 10x with the `-unfiltered-pl` argument during the generate-permit-list step. 

I first updated the docker image for Alevin-fry so it contains Alevin-fry:0.2.0 which can be pulled using `docker pull ghcr.io/alexslemonade/scpca-alevin-fry:latest`. In that update, there is a dependency on the most recent version of `libcxx`, requiring the install using the `-conda-forge` channel. 

I then added in the option to include the unfiltered barcode list from the 10X whitelist, which can be utilized by using the `--unfiltered` flag. When this option is used, the corresponding 10X barcode list for that sample is used, otherwise true cells are determined using the `--knee-distance` option. ~~I had tried to put the argument in a ternary operator and do something like ${params.unfiltered ? '-u' ${barcode_file} : '--knee-distance'} but nextflow doesn't like that. So I chose to do a conditional process based on the unfiltered parameter being true or not which works.~~

I also added in the resolution as a parameter. The default is set as full, but can be changed at the command line if we want to test other options.

### DIFF
--- a/images/alevinfry/Dockerfile
+++ b/images/alevinfry/Dockerfile
@@ -5,6 +5,6 @@ LABEL maintainer="ccdl@alexslemonade.org"
 RUN apt-get update && apt-get install -y procps
 
 # Install alevin-fry using bioconda
-RUN conda install -c bioconda alevin-fry 
+RUN conda install -c conda-forge -c bioconda alevin-fry=0.2.0
 
 WORKDIR /home
--- a/workflows/alevin-quant/run-alevin-fry.nf
+++ b/workflows/alevin-quant/run-alevin-fry.nf
@@ -9,6 +9,14 @@ params.annotation_dir = 'annotation'
 params.t2g = 'Homo_sapiens.GRCh38.103.spliced.tx2gene.tsv'
 params.mitolist = 'Homo_sapiens.GRCh38.103.mitogenes.txt'
 params.sketch = false // use sketch mode for mapping with flag `--sketch`
+params.resolution = 'full' //default resolution is full, can also use cr-like, cr-like-em, parsimony, and trivial
+
+params.barcode_dir = 's3://nextflow-ccdl-data/reference/10X/barcodes' 
+// 10X barcode files
+barcodes = ['10Xv2': '737K-august-2016.txt',
+            '10Xv3': '3M-february-2018.txt',
+            '10Xv3.1': '3M-february-2018.txt']
+params.unfiltered = false // use unfiltered list mode to include 10X barcode list with flat `--unfiltered`
 
 params.run_metafile = 's3://ccdl-scpca-data/sample_info/scpca-library-metadata.tsv'
 // run_ids are comma separated list to be parsed into a list of run ids,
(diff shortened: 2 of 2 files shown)


images/alevinfry/Dockerfile
```diff
@@ -5,6 +5,6 @@ LABEL maintainer="ccdl@alexslemonade.org"
 RUN apt-get update && apt-get install -y procps
 
 # Install alevin-fry using bioconda
-RUN conda install -c bioconda alevin-fry 
+RUN conda install -c conda-forge -c bioconda alevin-fry=0.2.0
```
jashapiro, line 8: > In that update, there is a dependency on the most recent version of libcxx, requiring the install using the -conda-forge channel.

I think this is fine, but I did wonder whether this could/should be installed with apt. As I write that, I actually think it may make more sense to install dependencies like this with `conda` (I assume there are other dependencies that are coming in here as well),  but we will probably want to make an accounting of what those all of those dependencies are at some point to keep things as stable/reproducible as possible. I don't think we are there yet, but we may want to do this when we finalize the package set we are using.

workflows/alevin-quant/run-alevin-fry.nf
```diff
@@ -9,6 +9,14 @@ params.annotation_dir = 'annotation'
 params.t2g = 'Homo_sapiens.GRCh38.103.spliced.tx2gene.tsv'
 params.mitolist = 'Homo_sapiens.GRCh38.103.mitogenes.txt'
 params.sketch = false // use sketch mode for mapping with flag `--sketch`
+params.resolution = 'full' //default resolution is full, can also use cr-like, cr-like-em, parsimony, and trivial
```
jashapiro, line 12: We might want to add an error check in the workflow to be sure that this value is valid before we get to the `quant_fry` process.

jashapiro (approved):
Looks good. I had a couple of minor optional suggestions, that are really more forward looking than required at this point. More things to keep in mind rather than things that need to happen now.

workflows/alevin-quant/run-alevin-fry.nf
allyhawkins: Definitely a good idea. For now I'm going to agree that this is something we can add in in the future if we end up doing more with this workflow. Right now, it will not run without that parameter being valid, but it would be good to check for any potential errors before running the other steps.

PR #88, opened 2021-05-12
Add per gene level benchmarking notebook

allyhawkins:
Related to #82, I have separated out the benchmarking analysis looking at metrics at a per gene level into a separate notebook. This is looking at the same samples that were imported into a `SingleCellExperiment` in #82 and now using the `scater::addPerFeatureQC` method to look at mean gene expression and percentage of cells each gene is expressed in across the pre-processing tools. 

We are using 4 single cell RNA seq samples and 2 single nuclei RNA seq samples, run on 4 different tools (cellranger, alevin, alevin-fry, and kallisto) with single nuclei RNA seq samples being run both on a cDNA and pre-mRNA index. In this notebook I chose to look at the correlation of mean gene expression and percent of cells a gene is detected in across each tool with cellranger for shared cells. I also looked at the percentage of genes that are overlapping between each tool and cellranger and the mean gene expression of those that overlap vs. those that do not overlap. 

I'm sure there are more things we can do, like looking at the overlap specifically of the most variable genes or highly expressed genes, but I wanted to start here because it does appear that most tools have fairly good overlap and high spearman correlation coefficients of > 0.9.


analysis/quantifier-comparisons/05-gene-level-benchmarking-metrics.Rmd
```diff
@@ -0,0 +1,398 @@
...
+```{r}
+ggplot(cell_detected_qc_common_cor, aes(x = cellranger, y = alevin)) +
+  geom_point(size = 0.5, alpha = 0.1) + 
+  facet_wrap(~ sample) + 
+  scale_x_log10() + 
+  scale_y_log10() + 
+  theme_classic()
+```
+Alevin doesn't look as correlated as kallisto does here, at least for 126 and 127. 
+We see almost a second group that has higher coverage in cellranger and is off the main diagonal.
```
jashapiro, line 284: This is interesting, and makes me want to see these plots for alevin-fry and sketch as well. Is this second group a common feature, or is it unique to alevin? I also think this is a place where implementing the `--cr-like` option in alevin-fry would be neat to look at in the future, if we go the alevin-fry route.

jashapiro (approved):
This looks good to me!

I can imagine there might be some additional analysis we might want to do looking at expression correlations within individual cells, but the earlier odd behavior of at least one sample in pairing cell_ids has me concerned, and I would like to resolve that before going too deep.

analysis/quantifier-comparisons/05-gene-level-benchmarking-metrics.Rmd
allyhawkins: It was a bit weird to see and is only unique to alevin. None of the alevin-fry combinations show it with all of them showing only one strong diagonal with no extra groups. I'm not entirely sure what the explanation is... but I do think as we narrow down tools, if we continue to use alevin-fry we should incorporate `--cr-like` as you said.

PR #89, opened 2021-05-18
Update Cellranger for mkfastq

jashapiro:
This PR updates the cellranger Dockerfile to also build bcl2fastq so that `cellranger mkfastq` can be run.

In order to build this, you now need to download bcl2fastq from Illumina as well as Cellranger, but I have not yet updated the documentation to reflect this fact, hence the draft status.

### DIFF
--- a/images/cellranger/Dockerfile
+++ b/images/cellranger/Dockerfile
@@ -1,7 +1,7 @@
 FROM ubuntu:20.04
 LABEL maintainer="ccdl@alexslemonade.org"
 
-RUN apt update && apt install -y build-essential libzip-dev
+RUN apt-get update && apt-get install -y build-essential libzip-dev
 
 # Install bcl2fastq
 WORKDIR /tmp
--- a/images/cellranger/Dockerfile
+++ b/images/cellranger/Dockerfile
@@ -1,4 +1,4 @@
-FROM centos:centos8
+FROM centos:centos7
 LABEL maintainer="ccdl@alexslemonade.org"
 
 RUN yum makecache && yum install -y unzip
--- a/images/cellranger/Dockerfile
+++ b/images/cellranger/Dockerfile
@@ -1,7 +1,5 @@
-FROM centos:centos7
+FROM centos:7
 LABEL maintainer="ccdl@alexslemonade.org"
-
-RUN yum makecache && yum install -y unzip
 
 # Install bcl2fastq
 WORKDIR /tmp
--- a/images/cellranger/Dockerfile
+++ b/images/cellranger/Dockerfile
@@ -1,9 +1,18 @@
 FROM ubuntu:20.04
 LABEL maintainer="ccdl@alexslemonade.org"
 
+RUN apt update && apt install -y build-essential libzip-dev
+
+# Install bcl2fastq
+WORKDIR /tmp
+COPY bcl2fastq2-v2.20.0.422-Source.tar.gz bcl2fastq2.tar.gz
+COPY install-bcl2fastq.sh install-bcl2fastq.sh
+RUN bash install-bcl2fastq.sh
+
+# Install cellranger
 WORKDIR /opt
-COPY cellranger-6.0.0.tar.gz cellranger-6.0.0.tar.gz
-RUN tar xzf cellranger-6.0.0.tar.gz && rm cellranger-6.0.0.tar.gz
-ENV PATH="/opt/cellranger-6.0.0:${PATH}"
+COPY cellranger-6.0.1.tar.gz cellranger-6.0.1.tar.gz
+RUN tar xzf cellranger-6.0.1.tar.gz && rm cellranger-6.0.1.tar.gz
+ENV PATH="/opt/cellranger-6.0.1:${PATH}"
 
 WORKDIR /home
--- a/images/cellranger/Dockerfile
+++ b/images/cellranger/Dockerfile
@@ -5,11 +5,9 @@ RUN yum makecache && yum install -y unzip
 
 # Install bcl2fastq
 WORKDIR /tmp
-COPY bcl2fastq2-v2-20-0-linux-x86-64.zip bcl2fastq2-v2-20-0-linux-x86-64.zip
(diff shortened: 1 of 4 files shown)


jashapiro:
I have now updated the README, but part of me wonders if we should switch this image from ubuntu to centos, so that we can use the illumina provided RPM file, which should make building much faster (compiling bcl2fast1 is not fast) and should also result in a smaller image. 

Probably a future problem.

images/cellranger/README.md
```diff
@@ -3,13 +3,17 @@ This folder contains a Dockerfile for the cellranger analysis.
...
+In order to build this image, the cellranger software and bcl2fastq source code components must be downloaded separately to comply with licensing, and should be placed in this folder (`images/cellranger`).
+- The current version of cellranger is `cellranger-6.0.1.tar.gz` and can be downloaded from [10X Genomics Website](https://support.10xgenomics.com/single-cell-gene-expression/software/downloads/6.0) after agreeing to their license terms.
+- The current version of bcl2fastq is 2.20.0 and can be downloaded from [Illumina](https://support.illumina.com/sequencing/sequencing_software/bcl2fastq-conversion-software.html) after logging into an Illumina account.
+  - As we are using an Ubuntu image, you will need to download the "Linux tarball" version of the software.
+  - Note that the file Illumina provides to download is a `.tar.zip` file that should contain the `bcl2fastq2-v2.20.0.422-Source.tar.gz` file specified in the `Dockerfile`. 
+  - Use `tar xzf bcl2fastq2-v2-20-0-tar.zip` to unpack the redundant packaging.
```
allyhawkins, line 11:
```suggestion
  - As we are using an Centos image, you will need to download the "Linux rpm" version of the software.
  - Note that the file Illumina provides to download is a `.zip` file that should contain the `bcl2fastq2-v2.20.0.422-Linux-x86_64.rpm` file specified in the `Dockerfile`. 
```
If we build with Centos than this will need to be changed to reflect grabbing the rpm file from illumina.

images/cellranger/Dockerfile
```diff
@@ -1,9 +1,18 @@
...
 LABEL maintainer="ccdl@alexslemonade.org"
 
+RUN apt-get update && apt-get install -y build-essential libzip-dev
+
+# Install bcl2fastq
+WORKDIR /tmp
+COPY bcl2fastq2-v2.20.0.422-Source.tar.gz bcl2fastq2.tar.gz
+COPY install-bcl2fastq.sh install-bcl2fastq.sh
+RUN bash install-bcl2fastq.sh
+
```
allyhawkins, line 11:
```suggestion
FROM centos:centos8
LABEL maintainer="ccdl@alexslemonade.org"

RUN yum makecache && yum install -y unzip

# Install bcl2fastq
WORKDIR /tmp
COPY bcl2fastq2-v2-20-0-linux-x86-64.zip bcl2fastq2-v2-20-0-linux-x86-64.zip
RUN unzip bcl2fastq2-v2-20-0-linux-x86-64.zip
RUN rpm -i bcl2fastq2-v2.20.0.422-Linux-x86_64.rpm \
&& rm bcl2fastq2-v2.20.0.422-Linux-x86_64.rpm && \
 rm bcl2fastq2-v2-20-0-linux-x86-64.zip
```
I tested this and it builds the image in ~ 2 minutes and I checked by going into the image interactively to make sure that bcl2fastq and cellranger was installed successfully. This isn't necessary since what you had worked, but this way everything is in one file and it's a little simpler.

allyhawkins (approved):
I think we definitely could have just left this as is, but your comment about using Centos had me curious and I wanted to test it out. So I gave it a try and using the rpm file and with Centos as a base image I was able to build it all within the Dockerfile wihout needing a separate install script and it was fairly fast taking only a few minutes. I checked and when you go into the image, the software is there. Either option could work, but maybe simpler to use the rpm file?

jashapiro:
> I think we definitely could have just left this as is, but your comment about using Centos had me curious and I wanted to test it out. So I gave it a try and using the rpm file and with Centos as a base image I was able to build it all within the Dockerfile wihout needing a separate install script and it was fairly fast taking only a few minutes. I checked and when you go into the image, the software is there. Either option could work, but maybe simpler to use the rpm file?

Let's do it! This takes way too long to build and I should have just switched rather than fighting with dependencies! Which centos did you use? (I think 7 is probably what we will want, because of some centos support changes that we don't need to get too far into now)

allyhawkins:
> Which centos did you use? (I think 7 is probably what we will want, because of some centos support changes that we don't need to get too far into now)

I originally used centos:8 as the base image but I just did centos:7 and it also worked fine so we should be good to go.

jashapiro:
Realized that this has been hanging out for a bit. I revised it to use the rpm file, but left the build script for posterity. Requesting rereview...

images/cellranger/install-bcl2fastq.sh
```diff
@@ -0,0 +1,29 @@
+#!/bin/bash
```
allyhawkins, line 1: I think we could delete this file since we no longer use it correct?

allyhawkins (commented):
Looks good to me, I just put in a note about the extra line installing `unzip` and then didn't think we needed the extra `install-bcl2fastq.sh` script anymore.

images/cellranger/install-bcl2fastq.sh
jashapiro: I decided to leave it, mostly for reference in case we decide we can't use the rpm in the future.

images/cellranger/Dockerfile
```diff
@@ -1,9 +1,17 @@
-FROM ubuntu:20.04
+FROM centos:centos7
 LABEL maintainer="ccdl@alexslemonade.org"
 
+RUN yum makecache && yum install -y unzip
```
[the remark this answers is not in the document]
jashapiro: probably not, will test.

images/cellranger/install-bcl2fastq.sh
allyhawkins: Makes sense 👍

PR #93, opened 2021-05-24
Update benchmarking notebooks to include correct data for SCPCR000006 alevin-fry-unfiltered-sketch

### LINKED ISSUE #90: Update SCPCR000006-alevin-fry-sketch-unfiltered run in current benchmarking analysis
After identifying that the near 0 correlation of the run of SCPCR000006 with alevin-fry-sketch using `--unfiltered-pl` with cellranger was due to an old version of the files being present during the sync from s3, we should update the benchmarking analysis that was done to have the correct data in it, so we have the correct benchmarking analysis for reference.

allyhawkins:
Closes #90. After discovering that in the current benchmarking analysis, one sample run, had very poor correlation with cellranger - SCPCR000006 run with alevin-fry` --sketch` and `--unfiltered-pl`. Investigation into the sample showed that the data had not been correctly copied over from s3 during import and the older version of the run during testing was being used instead. I have removed the previous versions of any data copies and rerun the notebooks and the correlation with cellranger is now as expected at 0.99. No code was updated, but the html files from the notebooks now contain the updated graphs and output for our future reference to reside in the `analysis/quantifier-comparisons` directory.


jashapiro (approved):
👍🏼

PR #94, opened 2021-05-24
Update alevin-fry docker to 0.3.0

allyhawkins:
In starting to run samples with the `splici` index and narrow in on testing alevin-fry, it was necessary to update the current docker image to alevin-fry 0.3.0. The new image has now been built and pushed to the git hub repository and can be pulled using `docker pull ghcr.io/alexslemonade/scpca-alevin-fry:0.3.0` or `docker pull ghcr.io/alexslemonade/scpca-alevin-fry:latest`.

### DIFF
--- a/images/alevinfry/Dockerfile
+++ b/images/alevinfry/Dockerfile
@@ -5,6 +5,6 @@ LABEL maintainer="ccdl@alexslemonade.org"
 RUN apt-get update && apt-get install -y procps
 
 # Install alevin-fry using bioconda
-RUN conda install -c conda-forge -c bioconda alevin-fry=0.2.0
+RUN conda install -c conda-forge -c bioconda alevin-fry=0.3.0
 
 WORKDIR /home


jashapiro (approved):
LGTM! Question (not for this PR): do we also need to update alevin in the workflow to a more recent version?

allyhawkins:
> Question (not for this PR): do we also need to update salmon in the workflow to a more recent version?

The requirements for alevin-fry are to have salmon >= 1.4.0 which we have, but there are 2 docker images associated with 1.4.0 and we have the older one in our workflow now. I can update to the newer one [1.4.0--h84f40af_1](https://quay.io/repository/biocontainers/salmon?tab=tags) in our workflow but I don't think it should make a huge difference.

jashapiro:
> > Question (not for this PR): do we also need to update salmon in the workflow to a more recent version?
> 
> The requirements for alevin-fry are to have salmon >= 1.4.0 which we have, but there are 2 docker images associated with 1.4.0 and we have the older one in our workflow now. I can update to the newer one [1.4.0--h84f40af_1](https://quay.io/repository/biocontainers/salmon?tab=tags) in our workflow but I don't think it should make a huge difference.

I don't think there is any reason to do that.. I actually noticed that salmon is now at 1.5.0, but there are no release notes I can find, so 🤷🏼 https://github.com/COMBINE-lab/salmon/releases/tag/v1.5.0

Issue #95, opened 2021-05-25
Adjust reference scripts to make files needed for splici index


allyhawkins:
Related to adding additional samples in #92 to be analyzed in alevin-fry using the `splici` index with the USA mode. I thought this would be an easy fix of just adding a few lines to generate the 3 column version of the `tx2gene.tsv` needed to run alevin-fry in USA mode. However, the steps outlined in the [alevin tutorial](https://combine-lab.github.io/alevin-fry-tutorials/2021/improving-txome-specificity/) are a bit different with some additional steps that we had not had prior to this. This results in a different `splici.fasta`, that is not exactly the same as the `spliced_intron.fasta` we originally were generating. We will need to add the necessary steps to generate these files to our current `make_pre_mrna_fasta.R` before processing any samples using the `splici` index.

allyhawkins:
The splici index has now been updated according to the alevin tutorial and is located in s3 here for alevin-fry:
`s3://nextflow-ccdl-data/reference/homo_sapiens/ensembl-103/salmon_index/spliced_intron_txome_k31`

PR #96, opened 2021-05-25
Update reference scripts to make splici.fa

allyhawkins:
To start addressing #95, we need to create the necessary fasta and tx2gene files for the `splici` index. I started by adding in just the addition of making the `tx2gene_3col.tsv` mapping that includes a third column containing information whether or not each transcript is categorized as spliced, unspliced, or ambiguous and then realized that the additional steps indicated in the [alevin-fry splici tutorial](https://combine-lab.github.io/alevin-fry-tutorials/2021/improving-txome-specificity/) altered the fasta output. Based on following the suggestions from the tutorial, they include a step to obtain unique sequences only before outputting the fasta. This was something we were not doing prior when creating our `spliced_intron.fa`. After looking at the output, it appears that about half of the sequences that are input into the `spliced_intron.fa` are duplicates and able to align to multiple transcripts. When looking at the sequences, they are duplicate, however the transcript ID's are unique. Additionally, all of the transcript ID's associated with duplicate reads are from intronic regions of transcripts. 

This could potentially be a problem when using the `spliced_intron.fa` with any of the tools that would throw out reads that are multi-mapping. Since we have decided to use this index with snRNA-seq samples with alevin-fry right now, this could impact how reads are assigned to each gene, as the full resolution is using an expectation maximization algorithm, while the `cr-like` mode assigns the read to the UMI with the highest count for that gene. However, it's unclear to me if we are losing sequences and potential information by adding this step. I reached out to Rob for some clarification and have not heard back yet. 

More information on alevin-fry quant can be found [here for reference](https://alevin-fry.readthedocs.io/en/latest/quant.html). 

Since we are interested in testing the `splici` index regardless, I went ahead and modified the scripts to generate the `splici.fa` in addition to the index files we already are generating. I added in a short comparison of the transcript ID's in the `compare_spliced_ensembl.Rmd` notebook we had, but think the real test is going to be does this alter the downstream results, which we will know after our next round of benchmarking. Happy to hear other thoughts on this!

### DIFF
--- a/analysis/reference-file-comparisons/compare_spliced_ensembl.Rmd
+++ b/analysis/reference-file-comparisons/compare_spliced_ensembl.Rmd
@@ -66,7 +66,6 @@ ensembl_fasta <- file.path(ref_dir, "fasta", "Homo_sapiens.GRCh38.txome.fa.gz")
 
 spliced_txome <- Biostrings::readDNAStringSet(spliced_fasta)
 ensembl_txome <- Biostrings::readDNAStringSet(ensembl_fasta)
-
 ```
 
 How long is each?
@@ -162,6 +161,43 @@ missing_tx_info %>%
 It certainly seems to be.
 
 Should we add the non-primary assembly to our references, or are we okay with these differences?
+
+
+## Splici vs. Spliced_Intron Fasta
+In generating the splici index, there is a step to obtain unique sequences only before writing the sequences to the fasta. 
+This results in the size of the splici.fa being much smaller than the spliced_intron.fa, suggesting that they are different and that the spliced_intron.fa contains duplicated sequences.
+
+```{r}
+# look at splici vs. spliced_intron fasta
+splici_fasta <- file.path(ref_dir, "fasta", "Homo_sapiens.GRCh38.103.splici.fa.gz")
+spliced_intron_fasta <- file.path(ref_dir, "fasta", "Homo_sapiens.GRCh38.103.spliced_intron.txome.fa.gz")
+
+splici_txome <- Biostrings::readDNAStringSet(splici_fasta)
+spliced_intron_txome <- Biostrings::readDNAStringSet(spliced_intron_fasta)
+```
+
+```{r}
+# get transcript ID's
+splici_fa_txs <- stringr::word(names(splici_txome),1)
+spliced_intron_fa_txs <- stringr::word(names(spliced_intron_txome),1)
+```
+
+```{r}
+# how many of the spliced_intron transcript ID's are unique? 
+unique_spliced_intron <- unique(spliced_intron_fa_txs)
+
+# what's the intersect between the spliced_intron and splici transcript ID's? 
+shared_splici_fa_tx <- intersect(splici_fa_txs, spliced_intron_fa_txs)
+
+glue::glue("
+  The splici transcriptome has: {length(unique_spliced_intron)} unique transcript IDs.
+  The splici transcriptome and spliced_intron transcriptome have: {length(shared_splici_fa_tx)} shared transcript IDs.
+  ")
+```
+
+
+This tells me that the spliced_intron.fasta contains duplicated sequences that map to multiple transcript ID's. 
+How 
 
 ## Session info
 
--- a/analysis/reference-file-comparisons/compare_spliced_ensembl.Rmd
+++ b/analysis/reference-file-comparisons/compare_spliced_ensembl.Rmd
@@ -178,8 +178,8 @@ spliced_intron_txome <- Biostrings::readDNAStringSet(spliced_intron_fasta)
 
 ```{r}
 # get transcript ID's
-splici_fa_txs <- stringr::word(names(splici_txome),1)
-spliced_intron_fa_txs <- stringr::word(names(spliced_intron_txome),1)
+splici_fa_txs <- names(splici_txome)
+spliced_intron_fa_txs <- names(spliced_intron_txome)
 ```
 
 ```{r}
@@ -194,10 +194,9 @@ glue::glue("
   The splici transcriptome and spliced_intron transcriptome have: {length(shared_splici_fa_tx)} shared transcript IDs.
   ")
 ```
+This tells us that the spliced_intron.fasta contains duplicated sequences that map to multiple transcript ID's. 
+This could likely be causing a problem for tools that throw out multi-mapping reads. 
 
-
-This tells me that the spliced_intron.fasta contains duplicated sequences that map to multiple transcript ID's. 
-How 
 
 ## Session info
 
--- a/workflows/rnaseq-ref-index/make_pre_mrna_fasta.R
+++ b/workflows/rnaseq-ref-index/make_pre_mrna_fasta.R
@@ -11,7 +11,7 @@
 # create a corresponding fasta, gtf, and transcript2gene mapping file 
 # for use with index generation for scRNA-seq 
 
-# This script is also used to generate a splici.txome.fa and splici.tx2gene_3col.tsv. 
+# This script is also used to generate the 3 column spliced_intron.tx2gene_3col.tsv needed for alevin-fry.
 
 # finally, a list of mitochondrial genes is output for QC
 
@@ -67,11 +67,8 @@ txome_tx2gene_file <- paste0(file_base, ".spliced.tx2gene.tsv")
 spliced_intron_fasta_file <- paste0(file_base, ".spliced_intron.txome.fa.gz")
 spliced_intron_gtf_file <- paste0(file_base, ".spliced_intron.txome.gtf")
 spliced_intron_tx2gene_file <- paste0(file_base, ".spliced_intron.tx2gene.tsv")
+spliced_intron_tx2gene_3col_file <- paste0(file_base, ".spliced_intron.tx2gene_3col.tsv")
 spliced_intron_metadata_file <- paste0(file_base, ".spliced_intron.metadata.tsv")
-
-## splici files 
-splici_fasta_file <- paste0(file_base, ".splici.txome.fa.gz")
-splici_tx2gene_3col_file <- paste0(file_base, "splici.tx2gene_3col.tsv")
 
 mito_file <- paste0(file_base, ".mitogenes.txt")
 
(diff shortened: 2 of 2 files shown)


workflows/rnaseq-ref-index/make_pre_mrna_fasta.R
```diff
@@ -123,14 +140,46 @@ eisaR::exportToGtf(
...
-# first make Tx2Gene for all spliced and intron sequences
+# create text file mapping transcript and intron identifiers to corresponding gene identifiers
+# make 2 column Tx2Gene for all spliced and intron sequences
 full_tx2gene <- eisaR::getTx2Gene(
   grl, filepath = spliced_intron_tx2gene
 )
 
-## need to write out to metadata 
+# trim grl used for splici
+splici_grl <- trim(grl)
```
jashapiro, line 150: What does trim do here? Is this the main difference remaining for splici and the previous grl (besides the later unique)?

The question here is if this is something we should just use for both...

workflows/rnaseq-ref-index/make_pre_mrna_fasta.R
```diff
@@ -123,14 +140,46 @@ eisaR::exportToGtf(
...
+splici_tx2gene_df <- eisaR::getTx2Gene(splici_grl)
+
+splici_tx2gene_df[, "status"] = sapply(
+  strsplit(splici_tx2gene_df$transcript_id, "-"), function(x) 
+    if(length(x) == 2){"U"} else {"S"}
+)
+
+splici_tx2gene_df[, "gene_id"] = sapply(
+  strsplit(splici_tx2gene_df$gene_id, "-"), function(x) x[1]
+)
```
jashapiro, line 177: I think I would rather see these as tidyverse commands for clarity. 

For example, I am trying to understand teh status column here, and I think the idea is that if there is a `-` in the transcript id that indicates it is unspliced, but that is *so* not how I would write it...

```suggestion
splici_tx2gene_df <- splici_tx2gene_df  %>%
  mutate(status = ifelse(str_count(transcript_id, '-') > 0, 'U', 'S')
         gene_id = str_extract(gene_id, "[^-]+")
```

Check that this does the same thing, of course, but I think it does?

jashapiro (commented):
I think this looks fine, but I have some questions about the redundancy of the two files. Should we just make one fastq file that we use for both versions of this? The only difference that I really see is the `trim()` step, and I don't really know what it is doing.

There were a couple places where "I don't really know what this is doing" occurred here, the other being the string splitting functions, where the use of base R obfuscates what the function is actually doing. In particular  something like

```
sapply(  strsplit(splici_tx2gene_df$transcript_id, "-"), function(x)  if(length(x) == 2){"U"} else {"S"})
``` 
makes it seems like we might split strings into any random number of parts, but only care about 2, which is super confusing to me. I think what is going on is that the dash occurs only in the introns, which then get marked as unspliced, but it took me way too long to figure that out.  I proposed a tidyverse alternative which I think is more clear, but I would probably add a comment as well.

analysis/reference-file-comparisons/compare_spliced_ensembl.Rmd
```diff
@@ -163,6 +162,42 @@ It certainly seems to be.
...
+# what's the intersect between the spliced_intron and splici transcript ID's? 
+shared_splici_fa_tx <- intersect(splici_fa_txs, spliced_intron_fa_txs)
+
+glue::glue("
+  The splici transcriptome has: {length(unique_spliced_intron)} unique transcript IDs.
+  The splici transcriptome and spliced_intron transcriptome have: {length(shared_splici_fa_tx)} shared transcript IDs.
+  ")
+```
+This tells us that the spliced_intron.fasta contains duplicated sequences that map to multiple transcript ID's. 
+This could likely be causing a problem for tools that throw out multi-mapping reads.
```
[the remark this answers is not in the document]
allyhawkins: > I think this looks fine, but I have some questions about the redundancy of the two files. Should we just make one fastq file that we use for both versions of this? The only difference that I really see is the `trim()` step, and I don't really know what it is doing.

Yes, other than the `unique()` and `trim()` step, they are the exact same. I also was very much back and forth on if we can just use this for both pre-mRNA and as the splici index and just convert to the splici index moving forward. I _think_ the answer is yes but wanted to get someone else's opinion on this. 

You make a good point about salmon removing duplicate sequences, something I didn't realize before. 
(I will also make the changes to the functions to tidyverse, I should have done that originally when I was looking at the tutorial)

workflows/rnaseq-ref-index/make_pre_mrna_fasta.R
rob-p: Hi @allyhawkins and @jashapiro --- sorry for not checking in on this the past couple of days.  The point if `trim` is that it prevents an out of bounds error if the intron plus the chosen flank extends past the end of the contig.  We found that sometimes there are a handful (usually just one or two) genes that are close to the end of a chromosome, and if you choose too long a flank length, the GenomicRange returned will overhang the chromosome.  This in turn causes an indexing error and the whole extraction procedure to fail.  The call to `trim()` prevents this, clamping intervals to start no earlier than the chromosome start and end no later than its length.

workflows/rnaseq-ref-index/make_pre_mrna_fasta.R
rob-p: I would also prefer this ;).  Would you be willing to contribute your modifications back to the function we have for this in our repo (https://github.com/COMBINE-lab/usefulaf) ?

jashapiro: Sure! I can't wait to have that repo name on my list! 🤣

jashapiro: When updating, I remembered another function that I always forget about:`stringr::word()`. 

So I would even more prefer
```
gene_id = word(gene_id, 1, sep = '-')
```
to the `str_extract` version above.

allyhawkins:
Thanks so much for the comments @rob-p. 

@jashapiro I consolidated the steps so we only have 1 `spliced_intron.fa` (or splici.fa) but kept the naming consistent with how we had previously had it. So the overall changes to the script here is the addition of the `tx2gene_3col.tsv` output that we will need. 

Since we are now modifying and updating how we create our `spliced_intron` index, we shouldn't need to make any more code changes, but will need to just re-generate the index with this new fasta. I went ahead and created a subfolder within `reference/homo_sapiens/fasta/spliced_intron_no_trim` and moved the `spliced_intron.fa` that we had previously generated to that location. I am planning on doing the same for the index's for now.

analysis/reference-file-comparisons/compare_spliced_ensembl.Rmd
```diff
@@ -163,6 +162,42 @@ It certainly seems to be.
...
+```
+
+```{r}
+# how many of the spliced_intron transcript ID's are unique? 
+unique_spliced_intron <- unique(spliced_intron_fa_txs)
+
+# what's the intersect between the spliced_intron and splici transcript ID's? 
+shared_splici_fa_tx <- intersect(splici_fa_txs, spliced_intron_fa_txs)
+
+glue::glue("
```
jashapiro, line 192: I hadn't noticed you using `glue`. 🎉

workflows/rnaseq-ref-index/make_pre_mrna_fasta.R
```diff
@@ -93,45 +99,69 @@ if (!dir.exists(opt$annotation_output)) {
...
+# write out 2 column Tx2 gene mapping
+readr::write_tsv(splici_tx2gene_df, spliced_intron_tx2gene, col_names = FALSE)
+
+# make 3 column Tx2 gene needed for alevin-fry USA mode
+splici_tx2gene_df_3col <- splici_tx2gene_df  %>%
+  # add status column
+  # remove extra -I* on gene ID
+  mutate(gene_id = stringr::word(gene_id,1,sep = '-'),
+         #if transcript_id contains an added -I*, then it is usnpliced
+         status = ifelse(stringr::str_detect(transcript_id, '-'), 'U', 'S'))
```
jashapiro, line 158: Do we want to make this comment more explicit in the code?
```suggestion
         #if transcript_id contains an added -I*, then it is usnpliced
         status = ifelse(stringr::str_detect(transcript_id, '-I'), 'U', 'S'))
```

workflows/rnaseq-ref-index/make_pre_mrna_fasta.R
```diff
@@ -93,45 +99,69 @@ if (!dir.exists(opt$annotation_output)) {
...
-# extract spliced and intron sequences
-# genomic regions defined above
-seqs <- GenomicFeatures::extractTranscriptSeqs(
+
+# add the levels and lengths
+seqlevels(grl) <- seqlevels(genome)
+seqlengths(grl) <- seqlengths(genome)
+
+# trim grl used for splici
+splici_grl <- trim(grl)
```
jashapiro, line 122: I don't think there is any major harm the way this is, but if we never use the bare `grl` after this point (we are not keeping two version around anymore) it might make sense to revert this to just `grl` to save memory.

jashapiro (approved):
Looks good; just one possible tiny change where the comment doesn't quite match the code (but hardly matters) and one small potential memory saving...

workflows/rnaseq-ref-index/make_pre_mrna_fasta.R
allyhawkins: I thought about that too and assume that it doesn't make a difference since it has to have a -I or -I followed by a number to be considered unspliced, so that will still do the trick and will make the change.

workflows/rnaseq-ref-index/make_pre_mrna_fasta.R
allyhawkins: So we actually do still use the regular grl. When you trim it, you lose the metadata from the grl, very unclear why. And until the USA mode was released, I had been grabbing the metadata (which maps the spliced gene to the intron gene id) and using that to collapse the counts. I don't want to get rid of that step just yet since I haven't actually had the change to play around with using the USA output.

PR #97, opened 2021-06-01
Adjust alevin to use splici index and USA mode

### LINKED ISSUE #91: Make pre-mRNA index the default index for use with single-nuclei RNA seq samples
After discussion about the benchmarking analysis, we decided to default to using the pre-mRNA index as the default index with all single nuclei RNA seq samples for alevin-fry and use the `--include-introns` flag when running cellranger. 

The workflows for alevin-fry and cellranger should be updated accordingly so we no longer need to separate sample runs by cell or nucleus or have a parameter option at the command line.

allyhawkins:
Closes #91. This PR changes some of the parameters in the alevin-fry and cellranger workflows being used for benchmarking. 

For alevin-fry, I have done the following: 
-  Combined the previous parameters for `index_name` and `t2g` to one `index_type` parameter to indicate if the workflow should be run using settings for the single-cell (`cell`) or single-nucleus (`splici`) samples. The default is set to `cell` and will then assign the `index_name` and corresponding tx2gene mapping file. 
- Added a parameter for the `tx2gene_3col.tsv` file needed to run Alevin-fry in USA mode. This can only be used if the `index_name` is set to `splici` and the resolution is `cr-like`. Otherwise, the 2 column tx2gene file is still used. 
- The unfiltered parameter has now been changed to `params.filter = unfiltered` as the default. If that is changed to `knee` then the `--knee-distance` flag is used to filter true barcodes during the `generate-permit-list step`. Based on our discussion of switching to only using the unfiltered version moving forward, I chose to set this as default and keep `knee` as an option just in case moving forward. 

Originally, I was planning on using the splici index as the default index for all single-nuclei RNA-seq samples, but with the addition of testing the splici index on single-cell samples that became less straight forward. Instead, I included a map of index_names and tx2gene files based on the index_type (currently set as `cell` or `splici`) with the anticipation that if we do choose to use an index for single-cell vs. single-nuclei, we can use the `seq_unit` column of the metadata to set the index and tx2gene (similar to how the barcode list is assigned now). 

For Cellranger, I set the default for single-nuclei samples to run with the pre-mRNA index using the `--include-introns` option, removing it as a parameter. 

I also did think about adding in checks to make sure the input parameters were correct, specifically the `params.filter` and the `params.resolution` options, but didn't think that was necessary quite yet. Probably something we want to consider adding if we do choose alevin-fry as our permanent workflow.

### DIFF
--- a/workflows/alevin-quant/run-alevin-fry.nf
+++ b/workflows/alevin-quant/run-alevin-fry.nf
@@ -32,7 +32,7 @@ barcodes = ['10Xv2': '737K-august-2016.txt',
             '10Xv3.1': '3M-february-2018.txt']
 
 // supported single cell technologies
-tech_list = ['10Xv2', '10Xv3', '10Xv3.1'] 
+tech_list = barcodes.keySet()
 
 // generates RAD file using alevin
 process alevin{
--- a/workflows/alevin-quant/run-alevin-fry.nf
+++ b/workflows/alevin-quant/run-alevin-fry.nf
@@ -23,6 +23,8 @@ barcodes = ['10Xv2': '737K-august-2016.txt',
             '10Xv3': '3M-february-2018.txt',
             '10Xv3.1': '3M-february-2018.txt']
 
+params.filter = 'unfiltered' // default filtering strategy is to use unfiltered, other options include --knee-distance
+
 params.run_metafile = 's3://ccdl-scpca-data/sample_info/scpca-library-metadata.tsv'
 // run_ids are comma separated list to be parsed into a list of run ids,
 // or "All" to process all samples in the metadata file
@@ -53,7 +55,7 @@ process alevin{
     path run_dir
   script:
     // label the run directory by id, index, and mapping mode
-    run_dir = "${id}-${index}-${params.sketch ? 'sketch' : 'salign'}-${params.resolution}"
+    run_dir = "${id}-${index}-${params.sketch ? 'sketch' : 'salign'}-${params.resolution}-${params.filter == 'knee' ? 'knee' : ''"}
     // choose flag by technology
     tech_flag = ['10Xv2': '--chromium',
                  '10Xv3': '--chromiumV3',
@@ -94,7 +96,7 @@ process generate_permit{
       -i ${run_dir} \
       --expected-ori fw \
       -o ${run_dir} \
-      -u ${barcode_file}
+      ${params.filter == 'knee' ? '--knee-distance' :  "-u ${barcode_file}"}
     """
 }
 
--- a/workflows/alevin-quant/run-alevin-fry.nf
+++ b/workflows/alevin-quant/run-alevin-fry.nf
@@ -10,10 +10,12 @@ params.sketch = false // use sketch mode for mapping with flag `--sketch`
 params.resolution = 'full' //default resolution is full, can also use cr-like, cr-like-em, parsimony, and trivial
 
 index_names_map = ['cell': 'spliced_txome_k31',
-                   'nucleus': 'spliced_intron_txome_k31']
+                   'splici': 'spliced_intron_txome_k31']
 
 t2g_map = ['cell': 'Homo_sapiens.GRCh38.103.spliced.tx2gene.tsv',
-       'nucleus': 'Homo_sapiens.GRCh38.103.spliced_intron.tx2gene.tsv']
+       'splici': 'Homo_sapiens.GRCh38.103.spliced_intron.tx2gene.tsv']
+
+params.t2g_3col = 'Homo_sapiens.GRCh38.103.spliced_intron.tx2gene_3col.tsv'
 
 params.barcode_dir = 's3://nextflow-ccdl-data/reference/10X/barcodes' 
 // 10X barcode files
@@ -31,6 +33,8 @@ params.outdir = 's3://nextflow-ccdl-results/scpca/alevin-fry-unfiltered-quant'
 // build full paths
 params.index_path = "${params.ref_dir}/${params.index_dir}/${index_names_map[params.index_type]}"
 params.t2g_path = "${params.ref_dir}/${params.annotation_dir}/${t2g_map[params.index_type]}"
+params.t2g_3col_path = 
+"${params.ref_dir}/${params.annotation_dir}/${params.resolution == 'cr-like' && params.index_type == 'splici' ? params.t2g_3col : t2g_map[params.index_type]}"
 
 // supported single cell technologies
 tech_list = ['10Xv2', '10Xv3', '10Xv3.1'] 
@@ -160,5 +164,5 @@ workflow{
   // collate RAD files 
   collate_fry(generate_permit.out)
   // create gene x cell matrix
-  quant_fry(collate_fry.out, params.t2g_path)
+  quant_fry(collate_fry.out, params.t2g_3col_path)
 }
--- a/workflows/alevin-quant/run-alevin-fry.nf
+++ b/workflows/alevin-quant/run-alevin-fry.nf
@@ -4,44 +4,40 @@ nextflow.enable.dsl=2
 // run parameters
 params.ref_dir = 's3://nextflow-ccdl-data/reference/homo_sapiens/ensembl-103'
 params.index_dir = 'salmon_index'
-params.index_name = 'spliced_txome_k31'
 params.annotation_dir = 'annotation'
-params.t2g = 'Homo_sapiens.GRCh38.103.spliced.tx2gene.tsv'
-params.mitolist = 'Homo_sapiens.GRCh38.103.mitogenes.txt'
+params.index_type = 'cdna' // default index type is cdna
 params.sketch = false // use sketch mode for mapping with flag `--sketch`
 params.resolution = 'full' //default resolution is full, can also use cr-like, cr-like-em, parsimony, and trivial
 
-//index_names = ['cell': 'spliced_txome_k31',
-//               'nucleus': 'spliced_intron_txome_k31']
+index_names_map = ['cdna': 'spliced_txome_k31',
+                   'splici': 'spliced_intron_txome_k31']
 
-//t2g = ['cell': 'Homo_sapiens.GRCh38.103.spliced.tx2gene.tsv',
-//       'nucleus': 'Homo_sapiens.GRCh38.103.spliced_intron.tx2gene.tsv']
+t2g_map = ['cdna': 'Homo_sapiens.GRCh38.103.spliced.tx2gene.tsv',
+       'splici': 'Homo_sapiens.GRCh38.103.spliced_intron.tx2gene.tsv']
 
 params.barcode_dir = 's3://nextflow-ccdl-data/reference/10X/barcodes' 
 // 10X barcode files
 barcodes = ['10Xv2': '737K-august-2016.txt',
             '10Xv3': '3M-february-2018.txt',
             '10Xv3.1': '3M-february-2018.txt']
-params.unfiltered = false // use unfiltered list mode to include 10X barcode list with flat `--unfiltered`
 
 params.run_metafile = 's3://ccdl-scpca-data/sample_info/scpca-library-metadata.tsv'
 // run_ids are comma separated list to be parsed into a list of run ids,
 // or "All" to process all samples in the metadata file
 params.run_ids = "SCPCR000001,SCPCR000002"
-
-params.outdir = 's3://nextflow-ccdl-results/scpca/alevin-fry-quant'
+ 
+params.outdir = 's3://nextflow-ccdl-results/scpca/alevin-fry-unfiltered-quant'
 
 // build full paths
-params.index_path = "${params.ref_dir}/${params.index_dir}/${params.index_name}"
-params.t2g_path = "${params.ref_dir}/${params.annotation_dir}/${params.t2g}"
-params.mito_path = "${params.ref_dir}/${params.annotation_dir}/${params.mitolist}"
+params.index_path = "${params.ref_dir}/${params.index_dir}/${index_names_map[params.index_type]}"
+params.t2g_path = "${params.ref_dir}/${params.annotation_dir}/${t2g_map[params.index_type]}"
 
 // supported single cell technologies
 tech_list = ['10Xv2', '10Xv3', '10Xv3.1'] 
 
 // generates RAD file using alevin
 process alevin{
-  container 'quay.io/biocontainers/salmon:1.4.0--hf69c8f4_0'
+  container 'quay.io/biocontainers/salmon:1.4.0--h84f40af_1'
   label 'cpus_8'
   tag "${id}-${index}"
   publishDir "${params.outdir}"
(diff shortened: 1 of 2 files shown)


workflows/alevin-quant/run-alevin-fry.nf
```diff
@@ -88,7 +96,7 @@ process generate_permit{
       -i ${run_dir} \
       --expected-ori fw \
       -o ${run_dir} \
-      ${params.unfiltered ? "-u ${barcode_file}" : '--knee-distance'}
+      ${params.filter == 'knee' ? '--knee-distance' :  "-u ${barcode_file}"}
```
jashapiro, line 99: I think we might want to move the filtering parameter options to the top of the script as a mapping/dictionary, as we have it in the `alevin` process. The reason is that we might want more flexibility if more filtering options come later. I am also not a huge fan of ternary operators in the case where somebody might input an unexpected value.

workflows/alevin-quant/run-alevin-fry.nf
```diff
@@ -127,7 +135,8 @@ process quant_fry{
      --tg-map ${tx2gene} \
      --output-dir ${run_dir} \
      -r ${params.resolution} \
-     -t ${task.cpus}
+     -t ${task.cpus} \
+     ${params.resolution == 'cr-like' ? '--use-mtx' : ''}
```
jashapiro, line 139: We are using mtx here so that we can use the alevin USA-reading function, correct? I feel like it is odd to tie this to `cr-like` as we could use that for the "normal" index as well, not just for `splici`.

workflows/alevin-quant/run-alevin-fry.nf
```diff
@@ -157,5 +166,5 @@ workflow{
   // collate RAD files 
   collate_fry(generate_permit.out)
   // create gene x cell matrix
-  quant_fry(collate_fry.out, params.t2g_path)
+  quant_fry(collate_fry.out, params.t2g_3col_path)
```
jashapiro, line 169: I'm confused here, which I think ends up related to the question at the top: `params.t2g_3col_path` is not always a 3 column file, correct? If so, maybe we should just keep the name as `t2g_path` for somewhat less confusion (since that is the ambiguous version, to me).

workflows/alevin-quant/run-alevin-fry.nf
```diff
@@ -47,7 +55,7 @@ process alevin{
     path run_dir
   script:
     // label the run directory by id, index, and mapping mode
-    run_dir = "${id}-${index}-${params.sketch ? 'sketch' : 'salign'}-${params.unfiltered ? 'unfiltered' : 'knee'}"
+    run_dir = "${id}-${index}-${params.sketch ? 'sketch' : 'salign'}-${params.resolution}-${params.filter == 'knee' ? 'knee' : ''"}"
```
jashapiro, line 58: I flipped the logic here to keep make this blank when unfiltered and add the filter name (and dash) only when it is something else. 

```suggestion
    run_dir = "${id}-${index}-${params.sketch ? 'sketch' : 'salign'}-${params.resolution}${params.filter == 'unfiltered' ? '' : "-${params.filter}" }"
```
For a bit more clarity (and shorter line length), we could break this up:
```suggestion
    run_dir = "${id}-${index}-${params.sketch ? 'sketch' : 'salign'}-${params.resolution}"
    if(params.filter != 'unfiltered'){
      run_dir += "-${params.filter}"
    }
```

workflows/alevin-quant/run-alevin-fry.nf
```diff
@@ -4,31 +4,39 @@ nextflow.enable.dsl=2
...
+params.outdir = 's3://nextflow-ccdl-results/scpca/alevin-fry-unfiltered-quant'
 
 // build full paths
-params.index_path = "${params.ref_dir}/${params.index_dir}/${params.index_name}"
-params.t2g_path = "${params.ref_dir}/${params.annotation_dir}/${params.t2g}"
-params.mito_path = "${params.ref_dir}/${params.annotation_dir}/${params.mitolist}"
+params.index_path = "${params.ref_dir}/${params.index_dir}/${index_names_map[params.index_type]}"
+params.t2g_path = "${params.ref_dir}/${params.annotation_dir}/${t2g_map[params.index_type]}"
+params.t2g_3col_path = 
+"${params.ref_dir}/${params.annotation_dir}/${params.resolution == 'cr-like' && params.index_type == 'splici' ? params.t2g_3col : t2g_map[params.index_type]}"
```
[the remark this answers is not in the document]
jashapiro: Looking at this again, it seems like we could put all of this in `t2g_path` as we don't (by my quick look) use that at all. 

More generally, I am not sure all of these paths have to be part of `params.` I did that earlier for some flexibility (you can specify a full path if you want to override the one built from components), but I am not sure it adds to long-term maintenence... Might be simpler to just make these regular variables that are created in the workflow itself.

I would also do something to make this shorter, like breaking out conditional logic to its own line.

workflows/alevin-quant/run-alevin-fry.nf
jashapiro: It looks also like this might be moot in the future? From https://combine-lab.github.io/alevin-fry-tutorials/2021/improving-txome-specificity/: 
> While presented as a stand-alone function above, we plan to add the functionality for loading and processing quantification estimates generated in USA-mode into fishpond / tximeta in the current development cycle.

jashapiro (commented):
I like the general direction of changes here, but I have some comments below on implementation. 

I think the main thing I am trying to understand is how the index and resolution options interact (or don't). It seems like if we use `splici`, we will always be using `cr-like` resolution? Or am I misunderstanding that? We might _also_ want to use `cr-like` with other tools, and perhaps that would become our new default...

Part of the motivation for `splici` also seemed to be that we might want to use it both for cell and nuclei (with the former to combat spurious transcripts).

workflows/alevin-quant/run-alevin-fry.nf
allyhawkins: Yes, I only added that in for right now and when I actually get a chance to play around with importing the quants files then we may be able to alter this. But my reasoning for tying it to `cr-like` is because that is the only resolution that has compatibility to be used with the USA mode and therefore would benefit from being processed using their USA functions. 

Are you saying that you think we should be using the `--use-mtx` flag for more than just the `cr-like` resolution? Or that we should be restricting it to `cr-like` and `splici` so it isn't also used for the cDNA index with `cr-like`? 

I also would love to be able to take it out in the future and have them all have the same output format, I just didn't want to do that without having a chance to actually look at both outputs.

allyhawkins:
Thanks for all the helpful comments! I went back through and I believe I have addressed all of your suggestions and comments now. The main edits I did were the following: 

1. I reorganized the workflow in general so all parameters are at the top, followed by any of the maps or lists that we use, then the processes, and lastly the workflow. 
2. I moved the index and tx2gene paths out of paramaters and into the workflow. Additionally, I omitted the use of a ternary operator to build the `tx2gene_3col_path` and am now using a separate if/else statement (or what is now being called the `t2g_quant_path` - I'll explain why I still think we need both of these below). 
3. I removed using the ternary operator for the choice in filtering options during `generate-permit-list` and now use a map within that process to assign the flag and barcode file if applicable. 

> I think the main thing I am trying to understand is how the index and resolution options interact (or don't). It seems like if we use `splici`, we will always be using `cr-like` resolution? Or am I misunderstanding that? We might _also_ want to use `cr-like` with other tools, and perhaps that would become our new default...

In regards to your confusion between use of `cr-like` and `splici` together, the only way that we can use the USA mode with single-cell RNA sequencing (and subsequently use the functions written by alevin-fry developers) is to use those two conditions simultaneously. However, we have been using the `splici` index with the `full` resolution and using our own functions to read in spliced + intronic counts and combining them specifically in the context of the single-nuclei samples. It was my understanding that if we are still testing `cr-like` vs. the `full` resolution (for both snRNA and scRNA) and the use of `splici` vs. `cDNA` for the single-cell samples we would want to have the option of being able to use any combination of those two. Once we have narrowed down a default, then we could combine it into one argument if we feel like that is necessary. I'm not sure I understand what you mean by also using `cr-like` with other tools? 

> I'm a little confused by the logic here. If we only use t2g_3col with the cr-like & splici settings, do we need to set t2g_3col_path to the other t2g file in the other cases?
> Looking at this again, it seems like we could put all of this in t2g_path as we don't (by my quick look) use that at all.

So this isn't actually true. For the scenario that we are doing `cr-like` and `splici`, we need both the 2 column tx2gene file and the 3 column file. The 2 column file is being used for the first alevin process during alignment and no matter what the resolution or index, we will always need to use the 2 column file for the first step. The 3 column file is what is needed only in the context of `cr-like` and `splici` and only during the last `alevin-fry quant` step. Otherwise, that step requires the 2 column tx2gene mapping. To try and clear this up a little in the actual code, I renamed it to be the `t2g_quant_path` and it is only set to the 3 column file with `cr-like` and `splici`, otherwise it is equivalent to the t2g_path used in the alevin step. If you have other suggestions on how to deal with this I am all ears!

workflows/alevin-quant/run-alevin-fry.nf
```diff
@@ -150,12 +156,25 @@ workflow{
...
+  index_prefix = "${params.ref_dir}/${params.annotation_dir}"
+  t2g_path = "${index_prefix}/${t2g_map[params.index_type]}"
+  
+  // if using splici and cr-like use the 3 column t2g file for alevin-fry quant
+  if(params.resolution == 'cr-like' && params.index_type == 'splici'){
+      t2g_quant_path = "${index_prefix}/${params.t2g_3col}"
+  } else{
+      t2g_quant_path = t2g_path
+  }
+
```
jashapiro, line 171: I would suggest moving these also to the start of the workflow, before the channels start.

workflows/alevin-quant/run-alevin-fry.nf
jashapiro: > Are you saying that you think we should be using the --use-mtx flag for more than just the cr-like resolution? Or that we should be restricting it to cr-like and splici so it isn't also used for the cDNA index with cr-like?

I think I was saying the latter, as I am pretty sure that we would be fine to read in cDNA/cr-like data with current scripts. 

I went back to read the example more carefully, and my understanding is that this should probably be conditional on the `tx2gene` file rather than the resolution mode alone? If the `tx2gene` file is three columns, that is the trigger for USA mode (I missed that point before), which then requires the mtx output so we can read the matrix (now 3x times longer) "manually" rather than with `tximport`/`tximeta`.  Since the use of 3col is contingent on `cr-like` and `splici`, I would expect that to be condition here.

That said, if we have to use the custom read function for mtx data sometimes, we could modify our downstream scripts to use that function (and mtx) everywhere.  But we probably don't want to recalculate things for benchmarking, and long term (hopefully sooner rather than later) I might rather we were using tximeta/tximport...

workflows/alevin-quant/run-alevin-fry.nf
```diff
@@ -150,12 +156,25 @@ workflow{
...
+  alevin(reads_ch, index_path, t2g_path)
   // generate permit list from alignment 
   generate_permit(alevin.out, barcodes_ch)
   // collate RAD files 
   collate_fry(generate_permit.out)
   // create gene x cell matrix
-  quant_fry(collate_fry.out, params.t2g_path)
-}
+  quant_fry(collate_fry.out, t2g_quant_path)
+}
```
jashapiro, line 180: terminal newline
```suggestion
}

```

jashapiro (commented):
My main comment here is about the `--use-mtx` flag, but there are a few more minor things in there as well.

I am not feeling particularly strongly about the mtx decision (my comment probably reflects that), but I think it comes down to my preferences being to either use it _everywhere_ or _only when necessary_. I think the _everywhere_ option is off the table for now (benchmarking phase), since it might require reprocessing, so I would probably favor _only when necessary_, which would mean not using it for the cDNA/cr-like combination, which I think it is being used for in the current workflow logic.

jashapiro:
> So this isn't actually true. For the scenario that we are doing `cr-like` and `splici`, we need both the 2 column tx2gene file and the 3 column file. The 2 column file is being used for the first alevin process during alignment and no matter what the resolution or index, we will always need to use the 2 column file for the first step. The 3 column file is what is needed only in the context of `cr-like` and `splici` and only during the last `alevin-fry quant` step. Otherwise, that step requires the 2 column tx2gene mapping. To try and clear this up a little in the actual code, I renamed it to be the `t2g_quant_path` and it is only set to the 3 column file with `cr-like` and `splici`, otherwise it is equivalent to the t2g_path used in the alevin step. If you have other suggestions on how to deal with this I am all ears!

Thanks for this explanation! That cleared it up in my head nicely, and the code changes I think help as well.

allyhawkins:
> I went back to read the example more carefully, and my understanding is that this should probably be conditional on the tx2gene file rather than the resolution mode alone? If the tx2gene file is three columns, that is the trigger for USA mode (I missed that point before), which then requires the mtx output so we can read the matrix (now 3x times longer) "manually" rather than with tximport/tximeta. Since the use of 3col is contingent on cr-like and splici, I would expect that to be condition here.

You are very right about the mtx output being dependent on the use of the 3 column tx2gene mapping, since that is how it designates USA mode or not. I went ahead and defined the use of that parameter within the same if statement for use of the 3 column file to address that. Although it does sound like we can get rid of this in the future hopefully as I would also prefer to use tximport.

workflows/alevin-quant/run-alevin-fry.nf
```diff
@@ -34,6 +34,20 @@ barcodes = ['10Xv2': '737K-august-2016.txt',
...
+t2g_path = "${index_prefix}/${t2g_map[params.index_type]}"
+  
+// if using splici and cr-like use the 3 column t2g file for alevin-fry quant
+if(params.resolution == 'cr-like' && params.index_type == 'splici'){
+    t2g_quant_path = "${index_prefix}/${params.t2g_3col}"
+    use_mtx = true
+} else{
+    t2g_quant_path = t2g_path
+    use_mtx = false
+}
```
jashapiro, line 49: I like this, but now I am inspired to make it a param (this is a bad idea, ignore it): 

But just to say what it would look like, you would have `params.use_mtx = false` at the top, then override it only in first case here (leaving the second case to whatever the user specified).

If I thought there was a chance we would want this workflow to allow mtx output at request, I would say it was a good idea, but we don't so it probably isn't.

workflows/alevin-quant/run-alevin-fry.nf
```diff
@@ -32,7 +32,21 @@ barcodes = ['10Xv2': '737K-august-2016.txt',
             '10Xv3.1': '3M-february-2018.txt']
 
 // supported single cell technologies
-tech_list = ['10Xv2', '10Xv3', '10Xv3.1'] 
+tech_list = barcodes.keySet()
+
+// file paths
+index_path = "${params.ref_dir}/${params.index_dir}/${index_names_map[params.index_type]}"
+index_prefix = "${params.ref_dir}/${params.annotation_dir}"
+t2g_path = "${index_prefix}/${t2g_map[params.index_type]}"
```
jashapiro, line 40: I had actually meant to leave this in the workflow block, just higher up, but I think either place is fine.

jashapiro (approved):
LGTM!

PR #112, opened 2021-07-15
Update docker image for Alevin-fry to v0.4.0

### LINKED ISSUE #103: Update alevin-fry docker image
Alevin-fry is now on version 0.4.0 and previously we were using version 0.3.0. The docker image needs to be updated accordingly.

allyhawkins:
Closes #103. This PR updates the Dockerfile for alevin-fry to build the image with the most current version, 0.4.0, of alevin-fry. I've tested that the image contains the correct version by running the image interactively and testing `alevin-fry --version`. 

I have also pushed this to github and tested that it can be pulled using: 
`docker pull ghcr.io/alexslemonade/scpca-alevin-fry:latest` or 
`docker pull ghcr.io/alexslemonade/scpca-alevin-fry:0.4.0`

### DIFF
--- a/images/alevinfry/Dockerfile
+++ b/images/alevinfry/Dockerfile
@@ -5,6 +5,6 @@ LABEL maintainer="ccdl@alexslemonade.org"
 RUN apt-get update && apt-get install -y procps
 
 # Install alevin-fry using bioconda
-RUN conda install -c conda-forge -c bioconda alevin-fry=0.3.0
+RUN conda install -c conda-forge -c bioconda alevin-fry=0.4.0
 
 WORKDIR /home


jashapiro (approved):
Approving this, but we probably don't need it anymore. Alevin-fry is now available on biocontainers so we can just use

`quay.io/biocontainers/alevin-fry:0.4.0--h7d875b9_0`

For workflows in the future. 

Reference link for future updates: https://biocontainers.pro/tools/alevin-fry

rob-p:
Since the salmon and fry containers are different, any salmon related bumps are probably done on a separate branch, but I just want to suggest bumping to salmon 1.5.1 while you're upgrading fry.  The are some small but synergistic changes between the two tools in the latest version that leads to a little sensitivity boost (at leas in our testing on simulated data).

allyhawkins:
> Since the salmon and fry containers are different, any salmon related bumps are probably done on a separate branch, but I just want to suggest bumping to salmon 1.5.1 while you're upgrading fry.

Yup! We reference the container for salmon in our nextflow workflow, so that is a separate update that is happening and is on our list.

PR #116, opened 2021-07-21
Add genes detected comparison across pre-processing tools

### LINKED ISSUE #109: Identify which genes are lost between alevin-fry and cellranger
In benchmarking we have thus far looked at the number of genes detected/cell and the correlation of mean gene expression between alevin-fry and cellranger, but we should do a quick check on the genes that are missing between alevin-fry and cellranger. In this check, we should look to see if the genes lost are mostly low expressed genes or not and if they belong to specific pathways that we may be particularly relevant to pediatric cancer.

allyhawkins:
Closes #109. Here I did a quick comparison to look at what genes are detected when using either cellranger or alevin-fry (either with `cr-like` or `cr-like-em`). To generate the rowdata data frame that is used, I ran the output files from running alevin-fry through the `benchmarking_generate_qc_df.R` script. 

For the most part all of the genes that are able to be quantified are the same, with ~ 200 genes that are found in cellranger but not found in alevin-fry. When looking at the genes that are not identified in all tools, they have lower mean gene expression. I then did a quick pathway analysis to see what type of genes would be lost if we were to use alevin-fry over cellranger. For this, I chose to do over-representation analysis with the list of "lost" genes as the input target gene list and all genes identified in alevin-fry and cellranger as the background list. I wasn't quite sure if this was the best approach, but think it should be sufficient in terms of looking at any pathways that may be enriched in the lost gene list. However, the low gene expression and low % of cells the lost genes are detected in makes me less worried about the overall contribution they would have to the final output. I think we would be safe in using alevin-fry and not lose important information in comparison to cellranger. If we really wanted to test that theory we could go further and do clustering and look for marker genes and see if any of those change, but I don't know if that's entirely necessary.


jashapiro (commented):
I think I agree with the overall conclusions here, which is to say that I don't think there is too much to see that would change our plans. That said, I do think there is some room for clarifying a few things and making some of the plots more informative. 

The main things I was thinking about (mostly covered in the line comments) are these:
- I think that restricting the analysis to shared cells would be more fair/informative. We may well have some extra filtering that we do, since `emptyDrops` seems quite conservative, and perhaps those cells are responsible for a good fraction of the "extra genes"? I suppose we could have both, but if we do, we would probably want to make the plots a bit more compact, which brings me to:
- upset plots may be nicer than venn diagrams for this data. It might even make it possible to facet on experiment which could be nice to separate out where the extra sets are coming in.
- The expression analysis should probably be done on normalized (or at least log transformed) data.  Having slightly less skewed data will make the comparisons much easier to visualize, I expect. I don't expect any conclusions to change, of course!

analysis/quantifier-comparisons/08-genes-detected-comparison.Rmd
```diff
@@ -0,0 +1,405 @@
...
+  dplyr::mutate(gene_status = ifelse(gene_id %in% c(cellranger_cr_intersect, cellranger_cr_em_intersect, cr_cr_em_intersect), "shared", "unique")) 
+```
+
+
+```{r fig.height = 5, fig.width=10}
+# make a plot to look at the mean gene expression of shared vs. unique genes for each tool
+ggplot(mean_exp_df, aes(x = alevin_resolution, y = mean, fill = gene_status)) + 
+  geom_boxplot() + 
+  facet_grid(~sample) + 
+  coord_cartesian(ylim = c(0,2))
```
jashapiro, line 243: Now that we are looking at log expression, we probably want to zoom in more.. maybe to 1?

analysis/quantifier-comparisons/08-genes-detected-comparison.Rmd
```diff
@@ -0,0 +1,405 @@
...
+If we now look at the gene expression for those genes that are unique to each tool, how do they compare to the gene expression for those genes that are shared?
+
+
+```{r}
+# calculate mean expression of log transformed counts matrix 
+log_row_means <- function(sce){
+   rowMeans(log1p(counts(sce))) %>%
+    as.data.frame() %>%
+    tibble::rownames_to_column("gene_id") %>%
+    dplyr::rename("mean" = ".")
```
jashapiro, line 210: Being more explicit here, since this ends up in a plot.
```suggestion
    dplyr::rename("mean_log_expression" = ".")
```

analysis/quantifier-comparisons/08-genes-detected-comparison.Rmd
```diff
@@ -0,0 +1,405 @@
...
+mean_exp_df <- mean_exp_df %>%
+  dplyr::mutate(gene_status = ifelse(gene_id %in% c(cellranger_cr_intersect, cellranger_cr_em_intersect, cr_cr_em_intersect), "shared", "unique")) 
+```
+
+
+```{r fig.height = 5, fig.width=10}
+# make a plot to look at the mean gene expression of shared vs. unique genes for each tool
+ggplot(mean_exp_df, aes(x = alevin_resolution, y = mean, fill = gene_status)) + 
+  geom_boxplot() + 
+  facet_grid(~sample) +
```
jashapiro, line 242: Just because this ended up really small, maybe 
```suggestion
  facet_wrap(~sample, nrow = 2) + 
```

analysis/quantifier-comparisons/08-genes-detected-comparison.Rmd
```diff
@@ -0,0 +1,405 @@
...
+alevin_fry_tools <- c("splici_salign_cr-like", "splici_salign_cr-like-em")
+# merge coldata df with quant_info
+coldata_info_df <- coldata_df %>%
+  # rename tool to be either cellranger or alevin-fry
+  dplyr::mutate(tool = ifelse(tool %in% alevin_fry_tools, "alevin-fry", "cellranger")) %>%
+  dplyr::left_join(quant_info,
+                   by = c("tool" = "tool", 
+                          "quant_id" = "quant_dir")) %>%
+  # rename the not_alevin to cellranger for plotting purposes
+  dplyr::mutate(alevin_resolution = ifelse(alevin_resolution == "not_alevin", "cellranger", alevin_resolution))
```
jashapiro, line 66: You do this here, but then later we get "not_alevin" again. I was confused.

analysis/quantifier-comparisons/08-genes-detected-comparison.Rmd
```diff
@@ -0,0 +1,405 @@
...
+}
+```
+
+
+```{r}
+# filter all sces together 
+all_sces <- list(cr_like_sce,
+                 cr_like_em_sce,
+                 cellranger_sce)
+names(all_sces) <- c('cr-like', 'cr-like-em', 'not_alevin')
```
jashapiro, line 96: Any reason not to do this?
```suggestion
names(all_sces) <- c('cr-like', 'cr-like-em', 'cellranger')
```

analysis/quantifier-comparisons/08-genes-detected-comparison.Rmd
```diff
@@ -0,0 +1,405 @@
...
+# visualize overlap between gene sets based on tool
+rowdata_filtered[which(rowdata_filtered$alevin_resolution == "not_alevin"), "alevin_resolution"] <- "cellranger"
+
+gene_detect_df <- rowdata_filtered %>%
+  group_by(gene_id) %>%
+  dplyr::summarise(tools_detected = list(unique(alevin_resolution)))
+
+gene_detect_df %>%
+  dplyr::distinct(gene_id, .keep_all = TRUE) %>%
+  ggplot(aes(x = tools_detected)) +
```
jashapiro, line 189: I know people do it all the time, but as a style thing, I hate piping into ggplot. 
As I was rewriting this, I think the `distinct()` call is not needed? The group_by/summarise should have taken care of that, right?

```suggestion
gene_detect_df <- rowdata_filtered %>%
  group_by(gene_id) %>%
  dplyr::summarise(tools_detected = list(unique(alevin_resolution))) 

ggplot(gene_detect_df, aes(x = tools_detected)) +
```

analysis/quantifier-comparisons/08-genes-detected-comparison.Rmd
```diff
@@ -0,0 +1,405 @@
...
+  dplyr::left_join(quant_info)
+
+# join back with quant_info to get sample information 
+```
+
+
+```{r}
+# add a column to classify as shared or unique to that tool
+mean_exp_df <- mean_exp_df %>%
+  dplyr::mutate(gene_status = ifelse(gene_id %in% c(cellranger_cr_intersect, cellranger_cr_em_intersect, cr_cr_em_intersect), "shared", "unique"))
```
jashapiro, line 234: When I was looking at this plot, I spend a lot of time thinking about why the cellranger graphs were higher, and I think it is because what I was interpreting "shared" as was shared across all, but that is not what you mean here. The confusion is due to all those genes that are found by alevin but not cellranger, most of which are probably because of annotation list differences. I think cellranger is more conservative it what it includes as a "gene" before even mapping. That could explain a lot of the difference. In any case, I think we can make this more explicit and informative with the following change (though I am not sold on these specific labels, and the code is not tested!):
```suggestion
mean_exp_df <- mean_exp_df %>%
  dplyr::mutate(gene_status = dplyr::case_when(
  gene_id %in% intersect(cellranger_cr_intersect, cellranger_cr_em_intersect) ~ "common",
  gene_id %in% c(cellranger_cr_intersect, cellranger_cr_em_intersect, cr_cr_em_intersect) ~ "shared", 
  TRUE ~ "unique")) 
```

jashapiro (commented):
This looks good, and the upset plot is much nicer to look at!

I think a lot of the differences we are seeing (I mention this in a comment below) are due to the fact that CR and alevin are simply looking at different gene sets. If we provided only the genes that cellranger has in its index to alevin, we would probably get a larger overlap fraction and more similar expression patterns. One way to deal with that is my suggestion below to look at the fully overlapping sets, but the other thing we could do is right at the start to restrict the analysis to only the common genes between the two _before_ filtering for expression > 0. 

(This assumes that cellranger is not filtering the gene list based on expression at all as part of its pipeline and will always output a counts matrix with the same set of genes. I would hope this is the case, as the alternative would make analysis so much harder. I know that this is what alevin/af does: it will report every gene you give it, regardless of expression.)

That said, if you separate out the "common to all" genes in the expression plot as I suggested below, I suspect that will provide the info I am looking for, either way.

And all that said, none of it _should_ affect questions about genes that we lose when not using cellranger, unless those genes happen not to be in the reference with alevin. Which is possible! We should check that!  I just looked, and we are using ensembl103 for alevin vs 100 for cellranger. There could certainly be differences due to that. 

Which is to say, we really need to check when we say something is "missing" that we can distinguish between missing due to the mapping and missing due to which genes are in the reference.

allyhawkins:
> One way to deal with that is my suggestion below to look at the fully overlapping sets, but the other thing we could do is right at the start to restrict the analysis to only the common genes between the two _before_ filtering for expression > 0.

Thanks for this suggestion and I think this makes complete sense! As the genes that are actually expressed will be sample dependent, while the ability to detect genes is what we are interested in. I removed the requirement of restricting genes to be expressed before making the comparisons and now you see only 40 genes that are not detected by alevin-fry that are detected by cellranger. When you do that though, there is no real clear way of comparing the genes in the different resolution modes, but I am less concerned about that. Also from what I can guess based on looking at a lot of cellranger output and always seeing around 19,000 genes I am pretty sure they also report all genes that are detected but I could be wrong. 

Also with this change, we now have the distinction of the gene being detected in all tools or being unique between cellranger and alevin-fry so have updated the gene expression plot as shared to mean found in all tools and unique to mean not found in all tools.

jashapiro:
This isn't quite what I meant... I still want to see the comparisons among mappers for the expression, but _only_ for the genes that are in both indexes. I went and checked the first two genes that are "missing" with CellRanger, and they are missing because they aren't in Ensembl 103, as I thought might be the case.The pages for [ENSG00000130723](https://useast.ensembl.org/Homo_sapiens/Gene/Idhistory?g=ENSG00000130723) and [ENSG00000263264](https://useast.ensembl.org/Homo_sapiens/Gene/Idhistory?g=ENSG00000263264) show that they were retired after v102, so this is not a tool difference, but an index difference. 

What I think we want to do is select only the genes that are in both indexes, regardless of expression, _then_ do the previous analysis where you looked at which are being expressed/detected in each sample and how that differs between the tools. It still isn't perfect (having the same reference version would be better!), but I don't think it is necessarily worth going back to redo all of the mapping.

allyhawkins:
Sorry about the mis interpretation of your previous comment @jashapiro     . I believe I have addressed what you were asking for and now have first restricted the analysis I was doing previously to only those genes that are present in the indices of both alevin-fry and cellranger. So before restricting by mean > 0 and detected > 0, I filtered the data frame containing the list of genes by those being found in both tools. Then I filtered on expression and then obtained the lists of genes found in each tool. I believe this should account for any changes in genes detected based on the index alone and should highlight any changes in expression of genes across the tools in the same set of samples. Regardless, I don't think we are losing much information and interestingly similar pathways are coming up in the gene sets that are unique to both cellranger and alevin-fry `cr-like-em`.

jashapiro (approved):
This looks good now! The plots definitely now look much more like what I would have expected in that there are no genes with even moderately high expression that are unique to any tool, with the possible exception of `cr-like-em`, which makes sense as that seems most likely to recover expression in genes where sequence similarity would otherwise have caused multi-mapped reads to be thrown out.

The fact that we see olfactory genes in both "unique" sets points to a marginal difference due to mapping strategies, and may be a residual effect of the reference differences. As the sets of OR genes changes, it seems reasonable to expect quantification changes in the common genes, as some reads might be now assigned to different sets of similar genes.

I think this analysis is good as is: the one other thing I might add if I were doing it would be to just look at the correlation between cellranger and cr-like-em with a linear model and identify any genes with large residuals that might fall somewhere other than at the bottom of the distribution.

That is, I would do something like `lm(cr_like_em$logmean_expr ~ cellranger$logmean_expr)` for each sample, then look at the residuals to see if there are other genes that are consistently over or undercounted, even if they are not showing up in the "unique to a method" set.

PR #119, opened 2021-07-23
Update salmon to 1.5.2 and alevin-fry to 0.4.1

### LINKED ISSUE #118: Update alevin-fry images to 0.4.1 (and maybe salmon?)
In response to some comments from us https://github.com/COMBINE-lab/alevin-fry/issues/22#issuecomment-885195051 `alevin-fry` now has better `.json` output files, so we should update. When we do that, some of the output files will change, but for the better! However, this will temporarily break scpcaTools (filing that issue separately).

We may want to wait for a coming version bump for salmon that will come in parallel, which adds back the `cmd_info.json` file for `--rad`  outputs.

One other note along with that is that we can remove the t2g file (`--tgMap`) from the `salmon alevin` call, as this is no longer needed with `--rad`

jashapiro:
Does what it says on the tin! As of Salmon version 1.5.0, the `--tgMap` option is not needed with `--rad`, so I removed it from the alevin workflow.

These version updates add much more informative output files for all stages of the workflow. Notably, though, the output file from `alevin quant` changes from `meta_info.json` to `quant.json`, which will require changes to scpcaTools (tracked in https://github.com/AlexsLemonade/scpcaTools/issues/24)

Closes #118

### DIFF
--- a/workflows/alevin-quant/run-alevin-fry.nf
+++ b/workflows/alevin-quant/run-alevin-fry.nf
@@ -89,7 +89,7 @@ process alevin{
 
 //generate permit list from RAD input 
 process generate_permit{
-  container 'quay.io/repository/biocontainers/alevin-fry:0.4.1--h7d875b9_0'
+  container 'quay.io/biocontainers/alevin-fry:0.4.1--h7d875b9_0'
   publishDir "${params.outdir}"
   input:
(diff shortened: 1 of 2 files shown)


allyhawkins (approved):
LGTM!

PR #120, opened 2021-07-27
CITE-seq workflo

jashapiro:
This PR adds a workflow to perform CITE-seq mapping and quantification with `salmon alevin`/`alevin-fry`. This workflow does not yet combine these data with the scRNAseq results; that will come in a separate PR, likely when incorporating these results into `scpca-nf`. For now, the goal is partly to have a set of output files that we can use to develop the steps required for creating SingleCellExperiment objects with both modalities of data.

This workflow is based largely on the existing `alevin-fry` workflow, but adds some elements and other changes. 

* Indexing is incorporated into this workflow, as the barcode index files may vary by sample. Indexing is very fast for these tiny files, but there are a couple of considerations to make it more efficient: 
  * The workflow first creates a channel that finds all unique feature barcode files from the requested sets, and indexes each only once.
  * The  feature index channel is then joined to the feature reads channel, matched by that index name, this is then sent for mapping.
  * The feature index also contains a dummy transcript 2 gene mapping file, which is needed by `alevin quant`. Since the index directory is small, we just pass along the whole index directory in the mapping output, though we could pass just the file. It should be a pretty easy change.

* I combined all `alevin-fry` steps into a single process. This should reduce some of the copying to and from S3, and should speed the workflow in general. We may well want to incorporate a similar change into RNAseq quantification workflow.

* In preparation for when the CITE-seq and RNA quants are merged, I made this workflow pass both `run_id` and `sample_id` along with the other required inputs and output. I think we will probably want to do something like this for the  `scpca-nf` workflow, so while we don't really use these here, this was a good place for a proof of concept.

I *think* we will be able to just combine RNA and CITE-seq by `sample_id` (using something similar to the `.combine` step here, but I will need to check that this works always. If it is not the case, we may need to add another column to the library info table which indicates which `run_id` contains the corresponding RNA for each CITE-seq or cell hash sample.

### DIFF
--- a/workflows/alevin-quant/run-af-feature.nf
+++ b/workflows/alevin-quant/run-af-feature.nf
@@ -1,0 +1,144 @@
+#!/usr/bin/env nextflow
+nextflow.enable.dsl=2
+
+// run parameters
+params.resolution = 'cr-like' //default resolution is cr-like, can also use full, cr-like-em, parsimony, and trivial
+params.barcode_dir = 's3://nextflow-ccdl-data/reference/10X/barcodes' 
+params.run_metafile = 's3://ccdl-scpca-data/sample_info/scpca-library-metadata.tsv'
+params.outdir = "s3://nextflow-ccdl-results/scpca/alevin-fry-features"
+
+// 10X barcode files
+barcodes = ['CITE-seq-10Xv2': '737K-august-2016.txt',
+            'CITE-seq-10Xv3': '3M-february-2018.txt',
+            'CITE-seq-10Xv3.1': '3M-february-2018.txt']
+// supported single cell technologies
+tech_list = barcodes.keySet()
+
+// run_ids are comma separated list to be parsed into a list of run ids,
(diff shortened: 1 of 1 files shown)


allyhawkins (approved):
Overall I think this looks great. I don't really have any big comments, and I do like how you combined the process for the post alignment alevin-fry steps. My only question is if we think there is any reason at all that we think we should be storing the index's for the feature barcodes for future use? At least right now while we are still in development? It looks like right now they aren't published and I feel like they are small enough that it might be worth it to publish that step and create a separate feature barcode index folder in the reference directory.

jashapiro:
I had thought about it, but that step is so very fast (less than 5 seconds) that it hardly seems worth it. I have code in there so we are only doing it for the number of indexes that exist, so it isn't like there are a bunch of extra copies hanging around.

If we did publish it, we would presumably want to have some code to look for the published version, which might actually make things more complex!

PR #124, opened 2021-08-02
Add linear regression analysis of mean gene expression across tools

### LINKED ISSUE #123: Add linear regression analysis to cr-like vs. cellranger gene comparison
Based on the comment below from #116, we should add in a quick linear regression analysis to the current gene comparison between alevin-fry `cr-like` and cellranger. This will help us identify if there are any genes that may be found in both sets but are consistently under or over represented in one of these tools. 

> I think this analysis is good as is: the one other thing I might add if I were doing it would be to just look at the correlation between cellranger and cr-like-em with a linear model and identify any genes with large residuals that might fall somewhere other than at the bottom of the distribution.

> That is, I would do something like lm(cr_like_em$logmean_expr ~ cellranger$logmean_expr) for each sample, then look at the residuals to see if there are other genes that are consistently over or undercounted, even if they are not showing up in the "unique to a method" set.

allyhawkins:
Closes #123. This PR adds a small section to the notebook looking at gene comparisons across tools. The comment in https://github.com/AlexsLemonade/alsf-scpca/pull/116#pullrequestreview-718525671 suggests that we should not only look at genes that are unique to each tool, but also look at how similar gene expression is across the tools. To do this, I have added in a linear regression analysis between cellranger and alevin-fry `cr-like` and `cr-like-em` and examined the residuals for each of the genes. 

If the tools are over or underestimating gene expression then the residuals will reflect that. I plotted the residuals for each gene in each sample in comparison to the mean gene expression and then labeled the genes that had an abs(residual) > 1 to identify genes that may be consistently over and under represented in alevin-fry in compared to cellranger. When doing this, I found that most of the genes were coming from ribosomal genes. I also found that there were much fewer outliers in the single-nuclei samples than the single-cell samples interestingly. 

I believe I have addressed the feedback in https://github.com/AlexsLemonade/alsf-scpca/pull/116#pullrequestreview-718525671, but please let me know if there were other comparisons that you had in mind.


analysis/quantifier-comparisons/08-genes-detected-comparison.Rmd
```diff
@@ -303,6 +306,66 @@ These genes that are found in cr-like-em include the following:
...
+                     names_from = alevin_resolution,
+                     values_from = mean_log_expression)
+
+# use a linear regression model to compare gene expression across tools 
+# grab the residuals for each gene and add to the original dataframe
+residuals_df <- mean_exp_df_spread %>%
+  dplyr::group_by(sample) %>%
+  dplyr::mutate(
+    cellranger_cr_like = lm(cellranger ~ `cr-like`)$residuals,
+    cellranger_cr_like_em = lm(cellranger ~ `cr-like-em`)$residuals
```
jashapiro, line 328: I feel like we should do the comparisons in the opposite direction here, since we are in general treating cellranger as the reference.
```suggestion
    cellranger_cr_like = lm(`cr-like` ~ cellranger )$residuals,
    cellranger_cr_like_em = lm(`cr-like-em` ~ cellranger)$residuals
```

analysis/quantifier-comparisons/08-genes-detected-comparison.Rmd
```diff
@@ -303,6 +306,66 @@ These genes that are found in cr-like-em include the following:
...
+  ) %>%
+  dplyr::left_join(mean_exp_df_spread) %>%
+  # condense all residuals for each tool comparison into one column for easy plotting
+  tidyr::pivot_longer(cols = c("cellranger_cr_like", "cellranger_cr_like_em"),
+                      names_to = "tool_comparisons", 
+                      values_to = "residuals") %>%
+  tidyr::pivot_longer(cols = c("cellranger", "cr-like", "cr-like-em"),
+                      names_to = "tool", 
+                      values_to = "mean_log_expression")
```
jashapiro, line 337: I think having the means for each here is overkill (and it doesn't make much sense to plot the residuals of a comparison between two variable against a third variable). I'd suggest just selecting one for plotting, but renaming first for clarity.
 ```suggestion
  dplyr::rename("cr_mean_log_expression" = "cellranger",
                "crlike_mean_log_expression" = "cr-like",
                "crlikeEM_mean_log_expression" = "cr-like-em")
```

File: analysis/quantifier-comparisons/08-genes-detected-comparison.Rmd [hunk 3]
```diff
@@ -303,6 +306,66 @@ These genes that are found in cr-like-em include the following:
...
+```
+
+```{r fig.height=10, fig.width=10}
+ggplot(residuals_df, aes(x = mean_log_expression, y = residuals)) + 
+  geom_point(size = 0.5) + 
+  facet_wrap(tool ~ sample, nrow = 3)+
```
jashapiro, line 343: I'm not sure why we are plotting by tool here, as the residuals should be associated with a comparison, not an individual tool. (If you look closely at the plot, you can see some genes are appearing more than once). I'm also using the suggestion above to only use the mean expression from CR.

```suggestion
ggplot(residuals_df, aes(x = cr_mean_log_expression, y = residuals)) + 
  geom_point(size = 0.5) + 
  facet_grid(sample ~ tool_comparisons)+ 
```

jashapiro (commented):
The overall analysis looks reasonable, but I had trouble interpreting some of the graphs & the table, as they were not quite what I was expecting to see. I made some suggestions below that I think should get closer to what I was expecting, but let me know if you have questions!

jashapiro (approved):
Looks good. 

I might quibble with some of the ways you were using of the words "over counted" & "under counted" with regard to residuals (on average residuals are by definition zero, and we don't know true values...). But the plots do illustrate that most of the change with the EM are at the very low end of expression, which I think is valuable information.

PR #126, opened 2021-08-15
Add scpcaTools to scpca-r image

### LINKED ISSUE #113: Add scpcaTools to scpca-r image
For future workflows, we should add the scpcaTools package and its dependencies to our `scpca-r` docker image.

We should be able to do with:
```
remotes::install_github("AlexsLemonade/scpcaTools")
```

jashapiro:
Closes #113

This modifies the docker image to include the latest version of the `scpcaTools` package (as of whenever the image are built, so this is not currently versioned).

I also added a couple other packages that might be useful; some are added just to be explicit.

I debated tying `scpcaTools` to a specific commit, but that seemed like overkill, at least for now. If we get to the point where we are tagging releases, it might be worth adding that to the `install_github` call, but that seems like a future issue.

### DIFF
--- a/images/scpca-r/Dockerfile
+++ b/images/scpca-r/Dockerfile
@@ -34,6 +34,7 @@ RUN Rscript -e "install.packages(c( \
 RUN Rscript -e "BiocManager::install(c( \
     'AnnotationHub', \
     'Biostrings', \
+    'bluster', \
     'BSgenome', \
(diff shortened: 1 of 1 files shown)


allyhawkins (approved):
This looks good to me.

jashapiro:
Image pushed!

PR #130, opened 2021-09-01
Add exploration of miQC filtering

allyhawkins:
This PR addresses the second points discussed in #105 of whether or not we can introduce usage of miQC as a secondary filtering step after `emptyDrops`. Here, I am comparing 2 of the single-cell samples and 2 of the single-nuclei samples that have been run through Alevin-fry with selective alignment to the `splici` index with the `--unfiltered-pl` filtering option and `cr-like-em` resolution (configurations chosen for processing). These samples have previously been filtered to remove any potential empty droplets using `emptyDrops` with `lower=200` as described in #128. I am keeping this in a draft stage right now because as we decide the thresholds to use for `emptyDrops`, I think we will want to update the input to this notebook. 

For the most part I chose to follow the [miQC vignette](https://github.com/greenelab/miQC/blob/main/vignettes/miQC.Rmd) and started by using the `plotMetrics()` function to inspect each of the samples and see if they fit the assumptions of the model. One of these assumptions is that they have cells with a range of mitochondrial contents, which when looking at the `plotMetrics()` output for the single-nuclei samples (SCPCR000118, SCPCR000119), it appears that this is not the case. 

I went ahead and applied miQC filtering using different posterior cutoffs and different non-linear mixture models in addition to the default settings and observed relatively similar results, regardless of the model that is used for the single-nuclei samples. I then compared the filtering to use of pre-determined thresholds to see how many cells are being removed and how filtering affects the distribution of UMIs and genes detected/cell. 

Some questions that I still have: 

- Are there other metrics that we should be measuring to compare filtering methods? 
- Would it be more beneficial to see the plots comparing distributions as density plots rather than boxplots (or is there another plot that would be helpful)?


jaclyn-taroni:
I have the same comment here as https://github.com/AlexsLemonade/alsf-scpca/pull/128#issuecomment-910245006.

analysis/quantifier-comparisons/10-miQC-filtering.Rmd
```diff
@@ -0,0 +1,272 @@
...
+To start we have to calculate the per cell statistics using `addPerCellQC()`. 
+```{r}
+# add per cell qc 
+cr_like_sce <- cr_like_sce %>%
+  purrr::map(scater::addPerCellQC, subsets = list(mito = mito_genes))
+```
+
+```{r}
+cr_like_sce %>%
+  purrr::map(miQC::plotMetrics)
```
jashapiro, line 67: Switching to iwalk prevents extraneous output, and we probably want some titles in these plots as well. I have not tested the code, but expecting (based on the source code) that `miQCplotMetrics()` returns a `ggplot2` object, this ought to work to add a title to each plot.
```suggestion
  purrr::iwalk(~miQC::plotMetrics(.x) + ggtitle(.y))
```

analysis/quantifier-comparisons/10-miQC-filtering.Rmd
```diff
@@ -0,0 +1,272 @@
...
+names(list_df) <- c("manual", "miQC", "pre_filtered")
+
+# one combined dataframe with coldata and filtering method as a column 
+filtered_df <- dplyr::bind_rows(list_df, .id = "filtering_method")
+```
+
+
+```{r}
+ggplot(filtered_df, aes(x = filtering_method, y = sum, fill = filtering_method)) +
+  geom_boxplot() +
```
jashapiro, line 233: I'm not finding these boxplots particularly informative. A density plots might be better, as they would show the whole distribution overlapped, which might allow some finer detail to be seen, but more generally, I'm just not sure that there is much information here beyond the numbers of cells. None of these are doing much that is dramatically different, aside from different cutoffs, and the scatter plots above already allow a reasonable comparison of where the manual cutoff would have been and some of these statistics.

analysis/quantifier-comparisons/10-miQC-filtering.Rmd
```diff
@@ -0,0 +1,272 @@
...
+  coord_cartesian(ylim = c(0,25000))
+```
+
+```{r}
+ggplot(filtered_df, aes(x = filtering_method, y = detected, fill = filtering_method)) +
+  geom_boxplot() +
+  facet_wrap(~ sample) + 
+  coord_cartesian(ylim = c(0,7500))
+```
+Based on the above plots, it appears that manual filtering is going to remove more of the cells with lower counts and genes detected than filtering with miQC is but keeping some of those may not be a bad thing.
```
jashapiro, line 244: I don't know what you mean here. Keeping some of which?

analysis/quantifier-comparisons/10-miQC-filtering.Rmd
```diff
@@ -0,0 +1,272 @@
...
+```{r}
+# test polynomial mixture model rather than linear
+purrr::map2(cr_like_sce, names(cr_like_sce), model_sce, model_type = "polynomial")
+```
+
+It looks like the change in model or different posterior cutoff doesn't have a drastic change in the cells that get included or excluded for this set of single-cell or single-nuclei samples. 
+
+## Filter Cells
+
+Now we can compare filtering using miQC vs. using the pre-determined cutoffs to see how the different filtering effects the overall number of cells that are removed.
```
jashapiro, line 139: I think it is worth mentioning here what these cutoffs are going to be. I can see from the code that you used 20% mito, but a reader might not see that easily, and the genes detected & count thresholds are not obvious either. 

My understanding of miQC is that it will keep low genes detected and counts, as the model doesn't really look for that, focusing on mitchondrial content (mediated by gene #). I would think we might want to use those even with miQC, so I feel like including them here might make for a bit of a biased comparison.  

Also, the total count threshold should largely be handled largely by emptyDrops: I don't think we want to use that as a hard cutoff after using emptyDrops specifically to avoid that hard cutoff!

jashapiro (commented):
I think this is a good start, but it gets to be a bit of a wall of results at points, so a bit more structure would be helpful to orient the reader a bit. (I agree with https://github.com/AlexsLemonade/alsf-scpca/pull/128#issuecomment-910245006)

I was thinking about how we might combine miQC with a threshold: One thought I had was that we could keep everything with less than 10% mito content (as a guess), or that passes miQC filtering. For cell data, I think this would basically mean that we would be using miQC, but would keep the high count/very low mito cells that were excluded by that model (see a comment below: these worry me a bit). For nuclei, this would revert to just a hard threshold, essentially ignoring most of miQC's rejections that might come from model assumption violations. Using a low mito threshold here makes sense to me as we do not expect nuclei to have very much if any mitochondrial gene content, as the mitochondria should be removed by the experimental procedure.

I'd definitely be curious to hear any thoughts from @arielah and @cgreene on these ideas, and congrats on the paper!

arielah:
Hi @jashapiro, you're right that there being high-count, low-mito cells being excluded is undesired behavior. In the devel version of the miQC package on bioconductor (should be released in November I think?), we've added a new parameter `keep_all_below_cutoff=TRUE` to the filtering functions make sure these cells are not excluded. You can either run the development version or do the cutoff you mentioned, which should accomplish roughly the same thing. Hope this is helpful!

allyhawkins:
Thanks for the feedback @jashapiro. I went ahead and updated the plots to remove the print outs of the lists along with the plots, make sure all the plots are named, removed the extraneous boxplots, and added in some commentary per your suggestions. 

It might be worth considering when we include our ["ccdl_suggests" column](https://github.com/AlexsLemonade/alsf-scpca/issues/134) that we integrate the posterior probability with also keeping cells that have above a high threshold of genes detected and low mito content as you suggested when we make those recommendations. So we don't lose those few cells that appear to get lost with the default settings of miQC that we agreed to use moving forward.

cgreene:
For miQC, an assumption is that you can successfully fit the mixture model. If your data are not a mixture (i.e., bad "cells" have already been filtered, you don't really have mitochondria, or the data are so clean that few cells are compromised) or if the compromised fraction is so small that you can't accurately estimate, miQC is probably a bad idea.

jashapiro:
I just want to add a note here for @arielah that SCPCR000126 in this case represents an interesting pathological case for the spline fit where the model intercept (which I think is used to determine which group to keep) happens to be lower in what should be the the compromised group, resulting in misclassification. I don't know how often people are using splines (they seem dangerous), but I thought it was worth a note. Not sure I have a good idea of a better solution... maybe using some kind of average of the lower portion of the fit rather than just the intercept point?

Screen Shot 2021-09-08 at 10 32 38 AM

analysis/quantifier-comparisons/10-miQC-filtering.Rmd
```diff
@@ -0,0 +1,234 @@
...
+```{r}
+# function to filter sce using miQC 
+miQC_filter <- function(sce, model_type = "linear", posterior_cutoff = 0.75){
+  sce_model <- miQC::mixtureModel(sce, model_type)
+  filtered_sce <- miQC::filterCells(sce, sce_model)
+}
+
+# filter using manual 
+manual_filter <- function(sce, 
+                          mito_cutoff = 10,
```
jashapiro, line 173: Just curious why you chose to bump this down to 10%? I thoguth the 20% cutoff made sense as a default (for single cell data at least).

analysis/quantifier-comparisons/10-miQC-filtering.Rmd
```diff
@@ -0,0 +1,234 @@
...
+```
+
+```{r}
+ggplot(combined_cell_num_df, aes(x = sample, y = number_cells, color = filtering_method)) + 
+  geom_point(size = 3) + 
+  theme(axis.text.x = element_text(angle = 90))
+```
+
+
+It looks like manual filtering is removing more cells than miQC for single-cell, but similar cell numbers are removed in single-nuclei samples with either manual or miQC.
```
jashapiro, line 227: Doesn't this just depend on the cutoffs we are using though? If we were using 20%, we would be getting a different relationship.

analysis/quantifier-comparisons/10-miQC-filtering.Rmd
jashapiro: Do we want to include a few different cutoffs (maybe w/ and without the `genes_detected_cutoff`) so we can see where miQC falls within those comparisons?

jashapiro (approved):
I think this is fine as an exploration, so I am approving it, but I'm still struggling a bit with how we might turn this into an action plan. 

What are our goals with these comparisons? Looking at the cell counts seems to be a pretty low-level metric and it is not clear how that relates to whether we are doing a good or bad job at removing compromised cells. 

What are the next steps after this? How should we think about these results when making recommendations on filtering?

I actually think if we want to evaluate the effects we may need to do something a bit more in depth where we look at PCA or UMAP results, and see if we are still getting clusters that seem to correspond to "bad cells". That is probably too much work at this stage, but I feel like it is something we might want to explore more, and at least point toward at the bottom of this document.

allyhawkins:
> What are our goals with these comparisons? Looking at the cell counts seems to be a pretty low-level metric and it is not clear how that relates to whether we are doing a good or bad job at removing compromised cells.
> 
> What are the next steps after this? How should we think about these results when making recommendations on filtering?

I would agree that only looking at the cell count is not the most informative and just gives us an idea of how many cells we are throwing out using each method. The goal with this PR was to see if using miQC was a viable option for the single-cell AND single-nuclei samples. The main focus being does applying the miQC mixture model to our samples work as expected in suggesting filtering of compromised cells. There are a few conclusions I would make after looking at the results: 
1. Using the default parameters for miQC with the linear mixture model gives the most consistent and reliable results where the majority of cells with low mito and high unique genes are classified as cells to keep. 
2. Cells from single-nuclei samples do not show as high a range of mito content and therefore the mixture model suggests a cutoff that leads to cells with a lower mito content than in single-cell being classified as compromised. This however is slightly expected considering single-nuclei samples should have little to no mito reads. 
3. From this we decided that we would not actually perform any further filtering ourselves, but rather apply miQC to each sample using the linear mixture model to obtain the posterior probability of a cell being compromised. Then we would create a `ccdl_suggests` column to identify which cells we would suggest filtering. 

The next steps to me are to identify the criteria that we need to recommend filtering. Now that we have settled on using miQC then I would say we need to identify if there are any other boundaries (such as combining a probability cutoff with a genes detected and mito cutoff) we should consider when creating the `ccdl_suggests` column. 

> I actually think if we want to evaluate the effects we may need to do something a bit more in depth where we look at PCA or UMAP results, and see if we are still getting clusters that seem to correspond to "bad cells". That is probably too much work at this stage, but I feel like it is something we might want to explore more, and at least point toward at the bottom of this document.

I think having additional PCA and UMAP comparisons would be helpful, and perhaps something to be added when we determine the criteria needed for suggested filtering in a future PR. I can add a few lines at the bottom of the PR to address these final conclusions and thoughts.

analysis/quantifier-comparisons/10-miQC-filtering.Rmd
allyhawkins: From a previous comment you made about moving the cutoff down to 10%, I had chosen to make that change (and based on our conversation in the scpca meeting that we wanted to keep everything below 10% mito). I think it would be fairly reasonable to add in a few different cutoffs to see where miQC falls, but I'm not entirely sure if that will change the decision to use miQC or not and think maybe that question falls under the next steps question of "how do we decide what additional filtering if any to include for the `ccdl_suggests` column"?

allyhawkins:
@jashapiro I went ahead and fixed the mito cutoff and added in comparisons of two different mito cutoffs and using just a mito cutoff with no genes detected to look at the number of cells that get removed. I also adjusted the last plot looking at the number of cells to be a bar plot to make it easier to compare across the various filtering settings within each sample. 

Additionally, I added some concluding remarks at the end of the file on some thoughts about what we learned here and next steps. Feel free to take a look at the [updated rendered html file](https://htmlpreview.github.io/?https://github.com/AlexsLemonade/alsf-scpca/blob/allyhawkins/miQC-filtering/analysis/quantifier-comparisons/10-miQC-filtering.nb.html) when you get a chance and let me know if you are okay with merging?

cgreene:
Just working through my inbox in stochastic order and saw the splines: definitely wouldn't recommend using splines. Also, I think the version of miQC associated with the paper has a flag to avoid throwing out cells in certain pathologic cases (i.e., never throw out a cell when one with fewer genes or higher mitochondrial content is preserved). If you're going to run miQC over many samples and may not manually inspect the results, this may be a thing you want to use.

PR #135, opened 2021-09-08
Modify cellranger workflow to allow for 5' samples

allyhawkins:
In order to run the 5' samples through Cell Ranger, I had to add in 5' to the list of acceptable technologies. Pretty straight forward as I added it to the `tech_list` and then was able to run Cell ranger with 5' libraries successfully.


jashapiro (approved):
LGTM!

PR #136, opened 2021-09-13
Add 5' support for alevin-fry workflow

### LINKED ISSUE #79: Support 5' libraries
In adding our first set of 5' libraries, we started to look at what current tools implement for mapping 5' reads. Quoting from https://github.com/AlexsLemonade/ScPCA-admin/pull/137#pullrequestreview-641296309:

> In looking at cellrangers description, they distinguish whether or not an assay is 5' or 3' based on the number of reads that are mapping in the antisense direction. If the fraction of reads from R2 mapping in the antisense direction is > 2X the reads mapping in the sense direction then they assume its in the 5' direction and proceed with the library like that. So in that case, rather than throwing away antisense reads, they would keep them, is what I'm interpreting from that statement.

We will need to investigate further. Once we decide on next steps, we will want to have workflows that implement those changes for 5' libraries. 

Currently the 5' libraries we have are designated as `10Xv2_5prime` in the technology column of the run table.

*Further notes*: 

Alevin looks like it can support 5' libraries by adjusting the library type argument to `-lISF` , as described here: https://github.com/COMBINE-lab/salmon/issues/439#issuecomment-547208043 It certainly seems that that 5' libraries are still stranded, so this should work, though the caveat in that comment makes me wonder a bit.

allyhawkins:
Closes #79. This PR adds in the ability to run 5' libraries using the alevin-fry workflow in this repo that we have been using for testing and benchmarking. I chose to keep things fairly simple here and added `10Xv2_5prime` to the `barcodes` and subsequently the `tech_list` to allow any sample with that technology to be run. 

According to https://github.com/COMBINE-lab/salmon/issues/439#issuecomment-547208043, we can modify Alevin-fry to work with 5' libraries by changing the  library type from `-lISR` to `-lISF`. To do this, I added in a ternary operator that will change the library type only when the 5' libraries are being used. I thought this would be the most straight forward way to do this for testing here, but I'm open to other suggestions on a better way to incorporate this change.

### DIFF
--- a/workflows/alevin-quant/run-alevin-fry.nf
+++ b/workflows/alevin-quant/run-alevin-fry.nf
@@ -81,7 +81,7 @@ process alevin{
     """
     mkdir -p ${run_dir}
     salmon alevin \
-      -l ${tech == '10Xv2_5prime' ? 'lISF' : 'lISR'} \
+      -l ${tech == '10Xv2_5prime' ? 'ISF' : 'ISR'} \
       ${tech_flag[tech]} \
       -1 ${read1} \
       -2 ${read2} \
--- a/workflows/alevin-quant/run-alevin-fry.nf
+++ b/workflows/alevin-quant/run-alevin-fry.nf
@@ -34,7 +34,8 @@ t2g_map = ['cdna': 'Homo_sapiens.GRCh38.103.spliced.tx2gene.tsv',
 // 10X barcode files
 barcodes = ['10Xv2': '737K-august-2016.txt',
             '10Xv3': '3M-february-2018.txt',
-            '10Xv3.1': '3M-february-2018.txt']
+            '10Xv3.1': '3M-february-2018.txt',
+            '10Xv2_5prime': '737K-august-2016.txt']
 
 // supported single cell technologies
 tech_list = barcodes.keySet()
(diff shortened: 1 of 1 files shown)


workflows/alevin-quant/run-alevin-fry.nf
```diff
@@ -34,7 +34,8 @@ t2g_map = ['cdna': 'Homo_sapiens.GRCh38.103.spliced.tx2gene.tsv',
 // 10X barcode files
 barcodes = ['10Xv2': '737K-august-2016.txt',
             '10Xv3': '3M-february-2018.txt',
-            '10Xv3.1': '3M-february-2018.txt']
+            '10Xv3.1': '3M-february-2018.txt',
+            '10Xv2_5prime': '737K-august-2016.txt']
```
jashapiro, line 38: Doesn't matter much here, but in the final workflow, I would probably want to keep all the v2s together.

workflows/alevin-quant/run-alevin-fry.nf
```diff
@@ -72,14 +73,15 @@ process alevin{
...
+                 '10Xv3.1': '--chromiumV3',
+                 '10Xv2_5prime': '--chromium']
     // run alevin like normal with the --rad flag 
     // creates output directory with RAD file needed for alevin-fry
     // uses sketch mode if --sketch was included at invocation
     """
     mkdir -p ${run_dir}
     salmon alevin \
-      -l ISR \
+      -l ${tech == '10Xv2_5prime' ? 'lISF' : 'lISR'} \
```
jashapiro, line 84: I think there are extra `l`s here?
```suggestion
      -l ${tech == '10Xv2_5prime' ? 'ISF' : 'ISR'} \
```

jashapiro (approved):
This looks fine for now. I might want to be a bit fancier in the production workflow, maybe moving the ternary operator out to make future changes a bit easier (if another 5prime library comes along, bascially), but I don't see any problem here, assuming it works as expected. 

I have not tested any of this myself!

PR #138, opened 2021-09-16
add orientation switch to af permit list step

jashapiro:
After helpful comments from @rob-p, I realized that we didn't do our orientation flip for 5' data in the correct place! 
https://github.com/AlexsLemonade/alsf-scpca/pull/137#issuecomment-920553327

This PR should fix that (or at least allow testing properly). I went with leaving the library orientation flag alone in the initial step, as that apparently should not matter, but we have to pass the tech along at least one more step, so I modified some inputs and outputs (in the scpca-nf workflow, this is already done).

For now, I went with the `rc` option for the 5' data, but we might still consider `both` as we do seem to have at a pretty significant number of forward reads from the previous analysis!

I should say that this change has not been directly tested as of filing this PR; I wanted to get the suggested changes posted before I will have a chance to do any testing. @ally-hawkins, if you have time to work on testing, that would be great!

### DIFF
--- a/workflows/alevin-quant/run-alevin-fry.nf
+++ b/workflows/alevin-quant/run-alevin-fry.nf
@@ -63,7 +63,7 @@ process alevin{
     tuple val(id), val(tech), path(read1), path(read2)
     path index
   output:
-    path run_dir
+    tuple val(id), val(tech), path(run_dir) 
   script:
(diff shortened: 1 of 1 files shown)


allyhawkins (approved):
All looks good and the results are looking much more like expected with these changes!

PR #140, opened 2021-09-17
Look at combinations of filtering with miQC, mito content, and genes detected

allyhawkins:
This PR addresses the last points brought up in #105 about whether or not to use miQC in combination with other filtering cutoffs to include in the `ccdl_suggests` column as part of the `colData`. Here, I'm adding to the previous notebook of exploring the use of miQC filtering. I am using samples that have been filtered with `emptyDrops` using `lower=200` and testing the effect of different filtering strategies on the sce object. 

After filtering I am showing plots of the distributions of UMI/cell and genes detected/cell colored by mitochondrial content along with PCA plots colored by both genes detected and mito content. I would anticipate that if a large population of compromised cells remained after filtering, that these cells with low genes detected and mito content would cluster together in the PCA plot. 

The conditions I am testing here include: 

1. miQC pass only 
2. miQC pass or below 10% mitochondrial reads 
3. (miQC pass or below 10% mitochondrial reads) and above 100 genes detected/cell
4. (miQC pass or below 10% mitochondrial reads)  and above 200 genes detected/cell
5. (miQC pass or below 10% mitochondrial reads) and above 500 genes detected/cell 

I'm putting this in draft form for now just to get some initial thoughts on the plots and to see if this is along the lines of what we are looking for when trying to decide what filtering strategy to go with. In looking at these plots I don't see any big differences between using miQC only and combining with other filtering in terms of the PCA plots. I do think combining miQC with a minimal mito threshold makes sense to me, but that being said, it doesn't appear to have a large effect on the dataset as  I can't see any large visual differences between each of these conditions.

### DIFF
--- a/analysis/quantifier-comparisons/10-miQC-filtering.Rmd
+++ b/analysis/quantifier-comparisons/10-miQC-filtering.Rmd
@@ -48,8 +48,9 @@ mito_genes <- mito_genes %>%
   unique()
 ```
 
+## MiQC Exploration
 
-## Plot Metrics 
+### Plot Metrics 
 
 Before we can do any modeling and perform any filtering, let's look at the distribution of mitochondrial fraction per cell and unique genes detected per cell. 
 The assumptions of miQC rely on having a distribution of mitochondrial reads and unique genes found. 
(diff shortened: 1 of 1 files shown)


allyhawkins:
Going through PR's and realizing this was still open. Going to close this for now and will re-open if we want to revisit any of this in the future.

PR #141, opened 2021-10-12
Add figures needed for FAQ - why did we use Alevin-fry?

allyhawkins:
Based on feedback received in AlexsLemonade/scpca-docs#21, I am breaking out the script used to create the figures needed for the FAQ: Why did we use Alevin-fry? I also am storing the figures here and will then include a permalink to the figures in the docs, rather than move them over to the `scpca-docs` repo. 

This script is hard-coded to take as input the results from previous benchmarking analysis that we've done with Alevin-fry using `cr-like` with selective alignment and Cell Ranger for two single-cell and single-nuclei samples. As part of the previous benchmarking analysis these sce objects were generated using `analysis/quantifier-comparisons/benchmarking_generate_qc_df.R`. 

I'm creating three plots to be used to compare Alevin-fry to Cell Ranger, two density plots of the distribution of UMI/cell and genes detected/cell and a scatter plot showing the correlation of mean gene expression between the two tools in each sample. After talking with Josh about the plots, we decided to switch to using a density plot to better compare the two distributions and then also show it on log-scale. Additionally, I changed the plots to be labelled with the library ID and what type of sample (cell or nucleus). 

I'm attaching the files here for easy review: 

UMI/cell comparison
Genes detected/cell comparison
Gene expression correlation


allyhawkins:
I went ahead and removed the color from the bars above the plots so that there's no more combining of plots and plots should now be consistent across all of them. I also went ahead and made the minor edits to the `aws s3 cp` statement for the metadata and removed the extra filtering step for the rowdata.  I am now only filtering for genes that are detected in > 5% of cells and then dropping any genes that are not found in both Alevin-fry and Cell Ranger using `drop_na()` when spreading the mean values into individual columns, rather than using the additional step.

analysis/docs-figures/fry-cellranger-faq-images.R
```diff
@@ -0,0 +1,168 @@
...
+  facet_wrap(~ plot_id) + 
+  scale_x_log10(labels = scales::label_number()) + 
+  scale_y_log10(labels = scales::label_number()) + 
+  labs(x = "Cell Ranger mean gene expression", y = "Alevin-fry mean gene expression") + 
+  theme_bw() + 
+  theme(axis.text = element_text(size = 6),
+        strip.background = element_rect())+
+  stat_cor(aes(label = ..r.label..), method = "pearson", size = 2)
+  
+ggsave(gene_exp_correlation)
```
jashapiro, line 168: Do we want to set the size here to match the other plots?

jashapiro (approved):
LGTM!

Issue #127, opened 2021-08-16
Explore genetic demultiplexing


jashapiro:
You may want to also add genetic demultiplexing as a complementary approach (or a ticket to have it as a possibility) if folks multiplexed samples. It may also reveal if there are free-floating nucleic acids around from one or more of the samples (genetic demultiplexing will fail on many cells with a mixture but cell hashing won't necessarily). I don't think this is a breaking issue (cell hashing _should_ be better unless the samples have some significant fraction of compromised tissue, in which case you'd just be filtering "cells" that ended up as a mixture with genetic demultiplexing.

_Originally posted by @cgreene in https://github.com/AlexsLemonade/scpca-nf/issues/9#issuecomment-899657426_

Reference paper (with a workflow we could adapt from): https://www.biorxiv.org/content/10.1101/2020.11.06.371963v3

jashapiro:
I am starting to work on this, and wanted to outline some of the steps... 

- First, we need a set of bulk samples for a reference data set. (In theory we can demux without that, but then assigning to a particular sample is problematic.
- We map each bulk sample (using STAR) to get a BAM file.
- The BAM files are then used with bcftools to call a set of SNPs, stored as a vcf (one per multiplexed set?)

Once we have the VCF in hand, we move to the single-cell portion
- The multiplexed single-cell samples are mapped with CellRanger (maybe we can substitute STARsolo here?)
- We call SNPs per cell with cellSNP/cellSNP-lite https://cellsnp-lite.readthedocs.io/en/latest/manual.html, using the VCF specific to the set of samples that have been multiplexed.
- Perform demultiplexing with Vireo. https://vireosnp.readthedocs.io/en/latest/

This leave a lot of detail to be filled in, but I hope to make progress by attacking each of these steps separately to start, then slowly stitching them into a more robust workflow.

jaclyn-taroni:
I'm going to go ahead and close this.

cgreene:
Oh - we found genetic demultiplexing to be better than cell hashing (more
recall, high concordance). If of interest, Ariel can share her results.

jaclyn-taroni:
[...]

PR #144, opened 2021-11-09
Add star indexing workflow

jashapiro:
As part of the genetic demulitplexing https://github.com/AlexsLemonade/alsf-scpca/issues/127, we need a STAR genome index. We will use this for bulk mapping to find SNPs.

This PR adds a workflow to create that index, putting the result with the other indices we have created at: s3://nextflow-ccdl-data/reference/homo_sapiens/ensembl-104/star_index/

Currently the index is based on the full gtf file; it might be a bit smaller if we reduced it to the gtf file that only includes the genes in the transcriptome gtf; I may test this.

Otherwise, this should be a pretty simple workflow. My only annoyance is that the input files had to be decompressed on disk :-(. 

The other question was the naming of the index output directory. I named it just with the Assembly, but I could add some suffix, like `.star_idx` if that seemed like something worth doing.

### DIFF
--- a/workflows/genetic-demux/index-star.nf
+++ b/workflows/genetic-demux/index-star.nf
@@ -12,7 +12,8 @@ STARCONTAINER = 'quay.io/biocontainers/star:2.7.9a--h9ee0642_0'
 process index_star{
   container STARCONTAINER
   publishDir "${params.ref_dir}/star_index", mode: 'copy'
-  label 'cpus_8'
+  memory "64.GB"
+  cpus "8"
   input:
     path ref_fasta
     path ref_gtf
@@ -32,7 +33,7 @@ process index_star{
       --genomeFastaFiles ${params.assembly}.fa \
       --sjdbGTFfile ${params.assembly}.gtf \
       --sjdbOverhang 100 \
-      --limitGenomeGenerateRAM 26000000000
+      --limitGenomeGenerateRAM 64000000000
     
     # clean up
     rm ${params.assembly}.fa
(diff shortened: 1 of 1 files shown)


jashapiro:
> Currently the index is based on the full gtf file; it might be a bit smaller if we reduced it to the gtf file that only includes the genes in the transcriptome gtf; I may test this.

Nope, doesn't make a difference. Unless there is another reason, I will stick with the default Ensembl file for simplicity.

workflows/genetic-demux/index-star.nf
```diff
@@ -0,0 +1,47 @@
...
+  publishDir "${params.ref_dir}/star_index", mode: 'copy'
+  memory "64.GB"
+  cpus "8"
+  input:
+    path ref_fasta
+    path ref_gtf
+  output:
+    path output_dir
+  script:
+    output_dir = "${params.assembly}"
```
allyhawkins, line 23:
```suggestion
    output_dir = "${params.assembly}.star_idx"
```
I do feel like having some sort of identifier that it's a star index even if it's in that folder could be helpful for the output. Only because we have so many different indices with the same assembly and if you were to ever use the name of the folder down the line I think it would be more helpful rather than harmful to be more descriptive here.

allyhawkins (approved):
This looks good to go to me. I did agree that we should have more information included in the output other than just the assembly to indicate exactly what's in the folder, just because of how many indices we have from this assembly already. Also really annoying that it doesn't take compressed files as inputs. Other than that this is good to go.

PR #145, opened 2021-11-11
Add bulk mapping with star

jashapiro:
As part of the genetic demulitplexing #127, we need to map bulk RNAseq data that we will use to discover SNPs in individual samples.

This PR adds a workflow to perform such mapping from bulk samples to create a BAM file and index it. Because the STAR image doesn't contain samtools, we have to do this in two steps, which is a bit annoying, but isn't terrible.

I am not saving/passing along the full STAR output folder, only the BAM file. The other files are small, so it would not be a burden to do that. I would probably in that case leave original file output name (fixed by STAR), and modify the samtools index script to take the directory and output it. But this seemed more direct, especially for a proof of concept.

### DIFF
--- a/workflows/genetic-demux/map-bulk-star.nf
+++ b/workflows/genetic-demux/map-bulk-star.nf
@@ -1,6 +1,9 @@
 #!/usr/bin/env nextflow
 nextflow.enable.dsl=2
 
+STARCONTAINER = 'quay.io/biocontainers/star:2.7.9a--h9ee0642_0'
+SAMTOOLSCONTAINER = 'quay.io/biocontainers/samtools:1.14--hb421002_0'
+
 // parameters
 params.ref  = 'Homo_sapiens.GRCh38.104'
 params.star_index = 's3://nextflow-ccdl-data/reference/homo_sapiens/ensembl-104/star_index/Homo_sapiens.GRCh38.104.star_idx'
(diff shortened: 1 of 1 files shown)


workflows/genetic-demux/map-bulk-star.nf
```diff
@@ -0,0 +1,92 @@
...
+    .map{[
+      run_id: it.scpca_run_id,
+      library_id: it.scpca_library_id,
+      sample_id: it.scpca_sample_id.replaceAll(";", "_"),
+      project_id: it.scpca_project_id?: "no_project",
+      submitter: it.submitter,
+      technology: it.technology,
+      seq_unit: it.seq_unit,
+      feature_barcode_file: it.feature_barcode_file,
+      feature_barcode_geom: it.feature_barcode_geom,
```
allyhawkins, line 74: Really minor, but I don't think you need these columns in the metadata for this workflow? 
```suggestion

```

workflows/genetic-demux/map-bulk-star.nf
```diff
@@ -0,0 +1,92 @@
...
+      s3_prefix: it.s3_prefix,
+    ]}
+    // only bulk samples
+    .filter{it.technology in bulk_techs} 
+    // use only the rows in the run_id list (run, library, or sample can match)
+    // or run by project or submitter if the project parameter is set
```
allyhawkins, line 80:
```suggestion

```
Just deleting this comment since you can't run by submitter or project ID here.

allyhawkins (approved):
This all looks good to me. That is really annoying that you have to do the indexing in a different process... I just had some minor comments on removing some parts of the `meta` tuple that we don't need and then deletion of an extraneous comment. Other than that looks good.

Issue #143, opened 2021-11-08
Explore spatial transcriptomics data 


allyhawkins:
Because we have a project that has spatial transcriptomics data, we will need to add a workflow to process the spatial transcriptomics data. Alevin-fry does provide a [tutorial on how to use Alevin-fry with spatial data](https://combine-lab.github.io/alevin-fry-tutorials/2021/af-spatial/) where they use Seurat to incorporate the image data with the Alevin-fry output. It appears that the alignment and quantification is performed the same as we would for single-cell or single-nuclei data with Alevin-fry and the data is loaded into R using the same functions, but they then convert it over to a Seurat object and add in the image data rather than keeping it as a `SingleCellExperiment` object. 

Another option could be to explore using the [`Spanial` package and the `SpatialExperiment` class](https://www.biorxiv.org/content/10.1101/2021.01.27.428431v2.full.pdf) which looks to be an extension of a SCE.

It might also be worth it at this stage to do a quick comparison of using [`spaceranger count`](https://support.10xgenomics.com/spatial-gene-expression/software/pipelines/latest/what-is-space-ranger) from 10X to compare the same metrics that we have compared prior (UMI/cell, genes detected/cell, mean gene expression) and look at the overlay of the image with the counts matrix. I don't foresee any issues with Alevin-fry alignment or quantification, but we should be sure that when we integrate the image it is being done correctly. 

Once we have identified the best method of performing alignment, quantification, image integration, and decided on how best to store the Spatial results, we can incorporate the workflow into `scpca-nf`.

jashapiro:
> Another option could be to explore using the [`Spanial` package and the `SpatialExperiment` class](https://www.biorxiv.org/content/10.1101/2021.01.27.428431v2.full.pdf) which looks to be an extension of a SCE.

Based on no real data, I think I would probably explore this route first. [The SpatialExperiment introduction vignette](https://bioconductor.org/packages/release/bioc/vignettes/SpatialExperiment/inst/doc/SpatialExperiment.html) seems well constructed. And how can you be upset when you see [their image example](https://bioconductor.org/packages/release/bioc/vignettes/SpatialExperiment/inst/doc/SpatialExperiment.html#adding-or-removing-images). It looks like the plotting functions come from [ggspavis](https://bioconductor.org/packages/release/bioc/html/ggspavis.html)

Oh, and there is an "Orchestrating" book: https://lmweber.org/OSTA-book/

[...]

PR #146, opened 2021-11-12
Add support for spatial transcriptomics libraries to Alevin-fry and Cell Ranger workflows

allyhawkins:
In order to start to address #143, we first needed to be able to process the Spatial libraries. I chose to modify both the alevin-fry and cellranger workflows in this repo to be able to accommodate running the spatial libraries so that next steps would include directly comparing the output from Alevin-fry to using the cell ranger equivalent, space ranger before choosing which workflow we would want to implement in `scpca-nf`. 

There were a few points to consider in doing this: 

- I started by following the [Alevin-fry tutorial on processing spatial data](https://combine-lab.github.io/alevin-fry-tutorials/2021/af-spatial/) and noticed that the actual use of Alevin-fry is no different between a single-cell library or using a spatial library. The main difference is that we actually don't have a barcode file to use during the `generate-permit-list` step for spatial libraries so we are forced to use the `knee-distance` filtering method in Alevin-fry. I even looked at the files that 10X provides when you search the slide serial numbers and the barcodes are not in the file, but only the X,Y coordinates for each spot with an extra code. I feel like there might be some other hidden spot on the 10X website where the barcodes live, but I was unable to find it so for now just tested this with running it using the knee filtering. 

- In order to use `spaceranger` I needed to add it into the docker image. Rather than make a brand new image, I chose to add it into the existing Cell Ranger image. Let me know if this is not the preferred method and I should make a separate image instead. 

- Right now we list all of the files for a sample in the `files` column of `scpca-library-metadata.tsv`, this includes grouping the image file with the fastq files for spatial libraries. I had to add a filtering step within the `getCRSamples` to only grab sample names from the fastq files, otherwise there will be an indexing error since not every file will have `.fastq.gz` appended and the indexing will fail. I also am grabbing the image file by just searching for the `.jpg` extension. An alternative to this would be to create a separate column in the metadata file for the image file from the start and then the filename can be taken from that column explicitly. We will need the image file for other steps downstream in processing with Alevin-fry as well so maybe this is worth doing? 

- While I was working on the cellranger workflow I also chose to modify it to use the same format that we have been using in `scpca-nf` of passing the metadata through the process.

### DIFF
--- a/images/cellranger/Dockerfile
+++ b/images/cellranger/Dockerfile
@@ -8,8 +8,13 @@ RUN rpm -i bcl2fastq2.rpm
 
 # Install cellranger
 WORKDIR /opt
-COPY cellranger-6.0.1.tar.gz cellranger-6.0.1.tar.gz
-RUN tar xzf cellranger-6.0.1.tar.gz && rm cellranger-6.0.1.tar.gz
-ENV PATH="/opt/cellranger-6.0.1:${PATH}"
+COPY cellranger-6.1.2.tar.gz cellranger-6.1.2.tar.gz
+RUN tar xzf cellranger-6.1.2.tar.gz && rm cellranger-6.1.2.tar.gz
+ENV PATH="/opt/cellranger-6.1.2:${PATH}"
+
+# Install spaceranger
+COPY spaceranger-1.3.1.tar.gz spaceranger-1.3.1.tar.gz
+RUN tar xzf spaceranger-1.3.1.tar.gz && rm spaceranger-1.3.1.tar.gz
+ENV PATH="/opt/spaceranger-1.3.1:${PATH}"
 
 WORKDIR /home
--- a/images/cellranger/Dockerfile
+++ b/images/cellranger/Dockerfile
@@ -12,9 +12,4 @@ COPY cellranger-6.1.2.tar.gz cellranger-6.1.2.tar.gz
 RUN tar xzf cellranger-6.1.2.tar.gz && rm cellranger-6.1.2.tar.gz
 ENV PATH="/opt/cellranger-6.1.2:${PATH}"
 
-# Install spaceranger
-COPY spaceranger-1.3.1.tar.gz spaceranger-1.3.1.tar.gz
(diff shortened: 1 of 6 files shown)


jashapiro (commented):
I know you asked about this earlier, and I said that putting both in the same docker image was fine, but I think I want to reconsider that. I think I had assumed that we might use both in a single process, but since that is not the case, I think it would be better to create two separate images. This keeps each smaller, and allows us to independently update them.

It took a while to find the barcode files, but they are hiding in`spaceranger-1.3.1/lib/python/cellranger/barcodes/` We will probably need to change the technology ids to `visium_v1` and `visium_v2`or something like that if we are to use and track those. I had thought about putting this off until later, but on some reflection, I think it might make sense to do at least the first step now: in `run-cellranger.nf` we could add visium tags to the technology lists, so that when we update the library file with those, it will still work. 

The  `run-alevin-fry.nf` changes I might pull into a separate PR, where we add the appropriate visium barcode files to S3, then use those for unfiltered processing. But I don't think that is necessarily important, as I expect the knee filtering should perform well for spatial data. 

Beyond that, the steps beyond the alevin-fry mapping will presumably need a lot of work if we are trying to run things without spaceranger at all; I'm not entirely sure we can do the work of aligning & pairing image files _without_ `spaceranger` so this may be a moot point for now (as in, our workflow may have to do spaceranger steps no matter what, though we might swap in alevin-fry quantification for consistency with other data).

Other than those two things (Docker images and technology names), I think this looks good.

allyhawkins:
Thanks for the helpful feedback @jashapiro. I went ahead and separated spaceranger and cellranger to be in two separate docker images, adding in a readme specific to the spaceranger image within that folder even though it is largely the same as cellranger. I then added the capability to process the `visium_v1` and `visium_v2` tags to `run-cellranger.nf` for now. I will include the updates to process unfiltered spatial data to Alevin-fry in a separate PR, so did not add the technology options there yet.

jashapiro (approved):
LGTM!

PR #147, opened 2021-11-16
Add starsolo workflow

jashapiro:
This PR adds a starsolo workflow to the genetic demultiplexing folder as part of #127 

While the end result is fairly simple, finding the exact sets of parameters so things would run consistently & without error was a bit of a challenge! In particular, the BAM sorting is finicky, but required for some of the SAM output fields that I think we will need for cellsnp later.

I separated the BAM output and starsolo quantification into different outputs, as I don't expect to need both downstream. I had considered skipping the quantification completely, but we do need it for some of the SAM output fields (maybe I can find the combination that doesn't require all of that, but the quantification is not the slow part, so I left it).

There is some redundancy here with the bulk workflow (bam indexing) but I left it for now, as I expect to do separate set of PRs where I integrate everything after individual steps work.

### DIFF
--- a/workflows/genetic-demux/map-sc-star.nf
+++ b/workflows/genetic-demux/map-sc-star.nf
@@ -54,7 +54,6 @@ process starsolo{
       --soloCellFilter EmptyDrops_CR \
       --outSAMtype BAM SortedByCoordinate \
       --outSAMattributes NH HI nM AS CR UR CB UB CY UY GX GN \
-      --outBAMsortingThreadN 2 \
       --limitBAMsortRAM 20000000000 \
       --runDirPerm All_RWX \
       --outFileNamePrefix ${output_dir}/ 
--- a/workflows/genetic-demux/map-sc-star.nf
+++ b/workflows/genetic-demux/map-sc-star.nf
@@ -52,10 +52,8 @@ process starsolo{
       --soloCBwhitelist ${barcode_file} \
       ${tech_flag[meta.technology]} \
       --soloCellFilter EmptyDrops_CR \
-      --outSAMtype BAM SortedByCoordinate \
+      --outSAMtype BAM Unsorted \
       --outSAMattributes NH HI nM AS CR UR CB UB CY UY GX GN \
-      --outBAMsortingThreadN 2 \
-      --limitBAMsortRAM 20000000000 \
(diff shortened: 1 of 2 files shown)


workflows/genetic-demux/map-sc-star.nf
```diff
@@ -0,0 +1,116 @@
+#!/usr/bin/env nextflow
+nextflow.enable.dsl=2
+
+STARCONTAINER = 'quay.io/biocontainers/star:2.7.9a--h9ee0642_0'
+SAMTOOLSCONTAINER = 'quay.io/biocontainers/samtools:1.14--hb421002_0'
```
allyhawkins, line 5: It might be worth pulling these out into the config file since you are using them here and in the bulk workflow.

workflows/genetic-demux/map-sc-star.nf
```diff
@@ -0,0 +1,116 @@
...
+  container STARCONTAINER
+  label 'bigdisk'
+  memory "32.GB"
+  cpus "8"
+  input:
+    tuple val(meta), path(read1), path(read2)
+    path star_index
+    path barcode_file
+  output:
+    tuple val(meta), path(output_dir), emit: starsolo_dir
```
allyhawkins, line 36: If you aren't using the first output, do you need to also output the metadata here since you are outputting it with the bam file in the second output? 
```suggestion
    path(output_dir), emit: starsolo_dir
```

allyhawkins (approved):
I think this all looks good. I just had two minor comments that you can feel free to incorporate or ignore, but everything else looks all good.

workflows/genetic-demux/map-sc-star.nf
jashapiro: Yes, eventually, but the mode for this repo has been so far for things to be mostly independent, so I wasn't doing that quite yet... Expect it in a later revision when I try to tie all the parts together!

workflows/genetic-demux/map-sc-star.nf
jashapiro: I actually want to keep this, with the thought that if we did want to proceed with the starsolo results into a different channel, we would still want the next step in the workflow to have the metadata. The cost of outputting it is tiny.

PR #149, opened 2021-11-18
Benchmarking of Spatial Transcriptomic libraries 

### LINKED ISSUE #143: Explore spatial transcriptomics data

allyhawkins:
Closes #143. Here I have added in a notebook that explores the output of running either Alevin-fry + Spaceranger or Spaceranger only on two spatial transcriptomic libraries. For the Alevin-fry quantification, I followed the guidance in [their tutorial](https://combine-lab.github.io/alevin-fry-tutorials/2021/af-spatial/). 

There are a few things that I looked at in this notebook: 

1. The use of a `SpatialExperiment` object vs. a `Seurat` object for storing the spatial data. In general, the `SpatialExperiment` object is fairly versatile and has similar functionality to the `SingleCellExperiment` with additional storage for spatial information. I also was able to make some plots using `ggspavis` and generally think that for keeping consistency with how we have been working with the single-cell and single-nuclei datasets, it would be good to continue with the `SpatialExperiment` objects. 
For loading in the data into R, I followed the recommendations set forth by the [OSTA Book](https://lmweber.org/OSTA-book/space-ranger-visium.html#import-outputs-into-r) and the [`SpatialExperiment` vignette](https://bioconductor.org/packages/release/bioc/vignettes/SpatialExperiment/inst/doc/SpatialExperiment.html).  
*One thing to note is that with the `SpatialExperiment` plots I couldn't figure out a way to plot it with the tissue underneath. I don't know if that's a dealbreaker or not or something that we will want down the line, but doing that might take some more finagling. 

2. I also compared the difference in using Alevin-fry for quantification or Spaceranger only for quantification, looking at mito content, UMI per spot, genes detected per spot, and then the actual distribution of these values across the spots in the tissue. I noticed that although the distributions are similar, when you look at the spatial plots, they don't quite lineup as expected which is slightly concerning (i.e. the distribution of the color of the spots is not the same for each sample when looking at each of the tools). 
I also looked at the correlation of mean gene expression and the overlap of genes detected and generally saw high correlation (other than one line of genes shifted off the diagonal) and high overlap. 

My overall conclusion that I gained after looking at these is that I wasn't sure it was worth the additional use of Alevin-fry and Spaceranger when we have to use Space ranger anyways for these samples, but let me know if reviewers came back with a different conclusion or would like to see other comparisons. 

[I'm attaching the html file of the rendered notebook here for reference. ](https://github.com/AlexsLemonade/alsf-scpca/files/7564307/12-spatial-transcriptomics.nb.html.zip)


analysis/quantifier-comparisons/12-spatial-transcriptomics.Rmd
```diff
@@ -0,0 +1,595 @@
...
+Additionally in this notebook we also explore the use of storing spatial data in a [`SpatialExperiment` object](https://bioconductor.org/packages/release/bioc/vignettes/SpatialExperiment/inst/doc/SpatialExperiment.html#3_Common_operations) or in a [`Seurat` object](https://satijalab.org/seurat/articles/spatial_vignette.html). 
+
+## Set Up
+
+```{r}
+library(magrittr)
+library(ggplot2)
+library(SingleCellExperiment)
+library(SpatialExperiment)
+library(ggspavis)
```
jashapiro, line 23: I'm noticing that `Seurat` is not loaded here, but functions later are called without `Seurat::`. I think I would probably leave Seurat unloaded if you can, but I might also remove `ggspavis` here. For following what is going on in an exploratory notebook, I find those prefixes very helpful.

analysis/quantifier-comparisons/12-spatial-transcriptomics.Rmd
```diff
@@ -0,0 +1,595 @@
...
+  
+  # read in image data
+  image_data <- SpatialExperiment::readImgData(spaceranger_dir,
+                            sample_id = sample_name)
+  
+  # read in spatial coordinates 
+  spatial_data_file <- file.path(spaceranger_dir, "tissue_positions_list.csv")
+  spatial_data <- readr::read_csv(spatial_data_file, col_names = c("barcode", "in_tissue", "array_row",
+                                                                   "array_col","pxl_row_in_fullres",
+                                                                   "pxl_col_in_fullres"))
```
jashapiro, line 109: Can we skip included the `col_names` here? Part of this is just a general preference, as this always seems brittle, but I am also trying to figure out why some of your images have flipped axes relative to one another. I doubt that this is the reason, but it is a distraction for comparisons.

analysis/quantifier-comparisons/12-spatial-transcriptomics.Rmd
```diff
@@ -0,0 +1,595 @@
...
+  spatial_data_subset <- spatial_data %>%
+    dplyr::filter(barcode %in% common_cells)
+  
+  # construct 'SpatialExperiment'
+  fry_spe <- SpatialExperiment(
+    assays = list(counts = assay(fry_sce_subset)),
+    colData = colData(fry_sce_subset), 
+    imgData = image_data,
+    spatialData = DataFrame(spatial_data_subset),
+    spatialCoordsNames = c("pxl_col_in_fullres", "pxl_row_in_fullres"),
```
jashapiro, line 126: Oh, I wonder if this is the place the coordinates get flipped. Usually rows before columns in R?

analysis/quantifier-comparisons/12-spatial-transcriptomics.Rmd
```diff
@@ -0,0 +1,595 @@
...
+  # find common cells between Alevin-fry and spaceranger 
+  spatial_data <- spatial_data %>%
+    dplyr::mutate(barcode = gsub("-1", "", barcode))
+  common_cells <- intersect(colnames(fry_sce), spatial_data$barcode)
+  
+  # subset fry_sce by common cells
+  fry_sce_subset <- fry_sce[,common_cells]
+  spatial_data_subset <- spatial_data %>%
+    dplyr::filter(barcode %in% common_cells)
+
```
jashapiro, line 119: I think this explains the difference in the plots later:  I expect `intersect()` is using the order from `fry_sce`, but then `filter()` is keeping the order of `spatial_data`, which may not be the same. (`fry_sce` is probably sorted by # of reads, `spatial_data` by position, which will explain the gradient you see in the fry data)

```suggestion
  common_cells <- intersect(colnames(fry_sce), spatial_data$barcode)
  
  # subset fry_sce by common cells
  fry_sce_subset <- fry_sce[,common_cells]
  spatial_data_subset <- spatial_data %>%
    dplyr::filter(barcode %in% common_cells)
  
```

analysis/quantifier-comparisons/12-spatial-transcriptomics.Rmd
allyhawkins: I think we want to add the column names here because this file doesn't have any column names and the later functions rely on the spatial data having specific names so we have to add them in somewhere. I do think the flip might be the order of the coordinates later on.

analysis/quantifier-comparisons/12-spatial-transcriptomics.Rmd
allyhawkins: This made me realize that this isn't the place where it's actually going wrong, it's in the plotting function, I'm plotting the columns as the x axis and the rows as the y axis, when it should probably be the other way around. This should also probably be flipped as well.

analysis/quantifier-comparisons/12-spatial-transcriptomics.Rmd
jashapiro: That makes sense. I might pull them out to a separate variable, just for clarity/formatting.

analysis/quantifier-comparisons/12-spatial-transcriptomics.Rmd
jashapiro: I don't think it is the plotting function, or the spaceranger is plots would be the same, as they use the same function. I mean, I don't know which is correct, but what you have looks fine to me. Column coordinate as x makes sense for plotting, even if it is opposite the standard (row, column) order.

allyhawkins:
Thanks for the help with this @jashapiro! One of your suggestions helped with why the pattern wasn't the same between spaceranger and the Alevin-fry + spaceranger version so that looks much better now. I'm still not entirely sure why they are flipped but something about when you are integrating the Alevin-fry results appears to flip the plot. I switched the coordinate names and then also tried plotting with both x/y axis but that obviously switches both versions so innately something is different about how the data is being stored in the two experiments. 

I also unloaded `ggspavis` and I believe I caught everything that uses that and `Seurat` and added the package name to the function, but let me know if I missed anything. I will also note one annoying thing about `Seurat` here is that to make the plot, it has a conflict with `SpatialExperiment` so I had to remove the `SpatialExperiment` package, and then make the plot before re-installing it again.

allyhawkins:
@jashapiro based on our offline discussion I went ahead and made some changes to these notebooks, breaking out the exploration between using a `SpatialExperiment` vs. `Seurat` into its own notebook and the benchmarking into its own notebook. In this PR, I kept in the notebook that compares the two objects and then also added in a functions into `benchmarking-functions/R` to read in Alevin-fry + Spaceranger output into an spe since we will be using that down the line. I am moving the additional benchmarking into its own notebook that I will file as a separate PR that is stacked on this one. 

I also added in the use of `plotVisium` to look at the plots with the image overlay as well for comparison, but overall, I think it still makes sense to continue to use the `SpatialExperiment` object rather than use `Seurat` here. Please let me know if there are any other comparisons or other things that I should explore for these objects that I missed.

analysis/quantifier-comparisons/12-spatial-transcriptomics-storage.Rmd
```diff
@@ -0,0 +1,206 @@
...
+
+```{r}
+# grab alevin fry and cellranger output
+sample_id <- c("SCPCR000372")
+
+aws_copy_samples(local_dir = s3_out,
+                 s3_dir = "s3://nextflow-ccdl-results/scpca",
+                 samples = sample_id,
+                 tools = c("alevin-fry-knee", "cellranger"))
+```
```
jashapiro, line 71: One thing that I noticed earlier: it might be nice to split these into separate chunks, with the download only happening if run manually, for easier rerunning of the notebook as a whole:
~~~suggestion
```{r}
# set samples to process
sample_id <- c("SCPCR000372")
```

```{r eval=FALSE}
# download alevin fry and cellranger output
aws_copy_samples(local_dir = s3_out,
                 s3_dir = "s3://nextflow-ccdl-results/scpca",
                 samples = sample_id,
                 tools = c("alevin-fry-knee", "cellranger"))
```
~~~

analysis/quantifier-comparisons/12-spatial-transcriptomics-storage.Rmd
```diff
@@ -0,0 +1,206 @@
...
+                     x_coord = "pxl_col_in_fullres",
+                     y_coord = "pxl_row_in_fullres")
+```
+
+```{r}
+Seurat::SpatialFeaturePlot(fry_seurat, features = "nCount_Spatial")
+```
+
+It looks like overall they both would work and that the `SpatialExperiment` will provide similar functionality to how we have things already set up for single-cell experiments. 
+One thing to note is that you cannot have `SpatialExperiment` installed at the same time that you attempt to do this plot.
```
jashapiro, line 206: I feel like I need more explanation here. Does this notebook not run if executed from top to bottom? I see the error about `SpatialImage` being found from Seurat, but is that affecting the running of this notebook directly?

If so, I think we need to figure out how to communicate that and/or include the installation/uninstallation in the notebook directly. 

Fundamentally, I think it is quite important to say directly that having both Seurat and SpatialExperiment installed (even if not loaded!) is a problem. In my mind, this is almost a dealbreaker.

jashapiro (commented):
Thanks for splitting this up! It seems much more manageable now. 

I had a few minor suggestions. The most important one is probably at the end: What exactly is the level of incompatiblity between `Seurat` and `SpatialExperiment`? If you really can't have both installed, that seems to me to be a potentially major problem. What happens besides the warnings when trying to make the plots? 

If it really is completely incompatible, it seems like it might be worth filing an issue in the SpatialExperiment repo. I don't know enough about R objects to suggest a solution (other than changing the object name), but it is something they should address!

allyhawkins:
@jashapiro I added some explanation about the level of incompatibility between `SpatialExperiment` and `Seurat` to the end of the notebook. You can run the notebook to completion except for the last chunk. The warnings will be produced but won't stop things from running. The last chunk where the plot is made using Seurat will not work though unless you remove `SpatialExperiment` and then reinstall `Seurat`. I found the solution on an issue that was filed in the Seurat repo where the `SpatialExperiment` developers also commented and added another comment about it still existing as an issue. I also tried to add in the removal of the package and reinstallation in the actual notebook and it did not seem to like that.

jashapiro:
>  I found the solution on an issue that was filed in the Seurat repo where the `SpatialExperiment` developers also commented and added another comment about it still existing as an issue. 

Can you add a link to that issue to the notebook? That will be helpful for somebody who comes to look at the notebook later, in part so we might be able to see if and when it is fixed.

allyhawkins:
> Can you add a link to that issue to the notebook? That will be helpful for somebody who comes to look at the notebook later, in part so we might be able to see if and when it is fixed.

Yes! Went ahead and added that in.

analysis/quantifier-comparisons/12-spatial-transcriptomics-storage.Rmd
```diff
@@ -0,0 +1,221 @@
...
+
+```{r}
+ggspavis::plotVisium(spe = fry_spe[,spatialData(fry_spe)$in_tissue == 1],
+                     x_coord = "pxl_col_in_fullres",
+                     y_coord = "pxl_row_in_fullres")
+```
+
+```{r}
+Seurat::SpatialFeaturePlot(fry_seurat, features = "nCount_Spatial")
+```
```
jashapiro, line 206: If I am understanding correctly, this chunk can not run with the rest of the notebook, so we should probably make it also not evaluate. I am a strong believer that a notebook should be able to run from top to bottom without errors. 

Because of this, we might want to separately include this image as an example, with the code as present here, but the note that it was not generated from this data set due to the incompatibility.

~~~suggestion
```{r eval=FALSE}
Seurat::SpatialFeaturePlot(fry_seurat, features = "nCount_Spatial")
```
~~~

analysis/quantifier-comparisons/12-spatial-transcriptomics-storage.Rmd
```diff
@@ -0,0 +1,221 @@
...
+head(colData(fry_spe))
+```
+
+You can still use the same functions to add in per cell metrics that we are using in `scpcaTools` now on a `SpatialExperiment`. 
+
+Let's specifically look at how each of them stores the data related to the spatial information.
+The `SpatialExperiment` has three places where it stores spatial information, `spatialData`, `spatialCoords`, and `imgData`. 
+
+```{r}
+head(spatialData(fry_spe))
```
jashapiro, line 157: Just a note that `spatialData` appears to be going away in a future version of this package. Nothing to do about it now, but worth keeping in mind. We might want to add a note/link here? https://github.com/drighelli/SpatialExperiment/issues/84

jashapiro (approved):
LGTM! I would just add another `eval=FALSE` tag, which might include separately adding the Seurat image. Then I would rerun the notebook top to bottom in a new session for a full refresh.

PR #151, opened 2021-11-22
Part 2 benchmarking spatial transcriptomics data

allyhawkins:
Here I am breaking out the second half of the notebook originally filed in #149 that address the comparison of quantifying spatial transcriptomics libraries with Alevin-fry + Spaceranger vs. Spaceranger alone into its own notebook. I have also paired down the analysis to only include comparison of one sample across both tools to decrease the complexity and make things easier to understand. Here I am looking at the same metrics that I looked at in the previous PR: UMI/ spot, genes detected/ spot, mean gene expression and overlapping genes. I also have included the plots to show the distribution of UMI/spot and genes detected/spot using both tools with and without the underlying image. 

In doing this, I added a few custom functions to the `benchmarking-functions/R` folder, specifically a function to read in the spaceranger output as a `SpatialExperiment` object and then two functions to grab the `colData` and `rowData` from an spe and convert to a `data.frame`. I'm hoping that by breaking these out here it will make the notebook less confusing. In the future I am anticipating switching some of these functions over from here to `scpcaTools`. 

The one function that I did not break out was the plotting function, since I was using that specifically in this notebook, but I can break them out if reviewers think that is necessary. I had to create two plotting functions, one for alevin-fry + spaceranger created spes and one for spaceranger only spes because of the bug that was identified in `SpatialExperiment::read10XVisium()` where the axes are flipped. 

Note that this is stacked on #149.


analysis/quantifier-comparisons/13-spatial-transcriptomics-benchmarking.Rmd
```diff
@@ -0,0 +1,411 @@
...
+                                     subsets = list(mito = mito_genes[mito_genes %in% rownames(.x)]))) %>%
+  purrr::map(scuttle::addPerFeatureQCMetrics)
+```
+
+After adding in the per spot QC metrics to both of the spe's, we want to extract the `colData` from each spe and create a data frame that we can use for plotting. 
+We will also need some information about each sample and how it was run, so we will create a sample metadata table, `sample_info_df` that will then be merged with the `colData`. 
+
+```{r}
+# create sample info dataframe to be joined with per spot dataframe later
+sample_info_df <- quant_info_table(data_dir= s3_out,
```
jashapiro, line 125: For clarity here, it might be worth printing this table. It is unclear what the contents will be within the notebook, as the function is elsewhere.

analysis/quantifier-comparisons/13-spatial-transcriptomics-benchmarking.Rmd
```diff
@@ -0,0 +1,411 @@
...
+                 tools = c("cellranger", "alevin-fry-knee"),
+                 samples = sample_id) %>%
+  # convert cellranger to spaceranger 
+  dplyr::mutate(tool = ifelse(tool == "cellranger", "spaceranger", tool))
+```
+
+When we convert the `colData` to a data frame we use the custom function, `spatial_coldata_to_df()` to do so and apply it to each spe in our list. 
+```{r}
+# join coldata dataframe with sample info
+coldata_df <- purrr::map_df(all_spe_list, spatial_coldata_to_df, .id = "tool") %>%
```
jashapiro, line 135: Just for consistency, we might want to use all `purrr` functions the same way:
```suggestion
coldata_df <- all_spe_list %>%
  purrr::map_df(spatial_coldata_to_df, .id = "tool") %>%
```

analysis/quantifier-comparisons/13-spatial-transcriptomics-benchmarking.Rmd
```diff
@@ -0,0 +1,411 @@
...
+  dplyr::left_join(sample_info_df,
+                   by = c("tool"))
+```
+
+Now we only want to filter our data frame to contain spots that are shared between both tools and those that are found to be overlapping with the tissue. 
+
+```{r}
+# identify shared spots only 
+spot_counts <- coldata_df %>%  
+  dplyr::count(spot_id, sample)
```
jashapiro, line 147: I think we should see what this result is? I would expect that we would get the same spots, but if we are not, how often is that happening?

analysis/quantifier-comparisons/13-spatial-transcriptomics-benchmarking.Rmd
```diff
@@ -0,0 +1,411 @@
...
+filter_spe <- function(spe){
+  spe <- spe[, spatialData(spe)$in_tissue == 1]
+}
+
+all_spe_filter <- all_spe_list %>%
+  purrr::map(filter_spe)
+```
+
+When we look at our results, we will also want to visualize them so we will make a custom function to plot the results.
+Because of the bug that is present in `SpatialExperiment::read10XVisium()` function, we will make two separate functions, one for spaceranger only spes and one for alevin-fry + spaceranger spes.
```
jashapiro, line 172: As I said earlier, I would rather work around this bug directly at read time (or avoid the function entirely).

analysis/quantifier-comparisons/13-spatial-transcriptomics-benchmarking.Rmd
```diff
@@ -0,0 +1,411 @@
...
+Generally it looks like both tools are fairly similar in the number of UMI/cell and genes detected/cell.
+
+### Per Gene QC Metrics
+
+Let's also look at the correlation of mean gene expression across shared genes. 
+We will first need to calculate the per feature QC on the filtered spes after removing spots not present in the tissue and then grab the `rowData` and combine into a data frame used for plotting. 
+
+```{r}
+all_spe_filter <- all_spe_filter %>%
+  purrr::map(scuttle::addPerFeatureQCMetrics)
```
jashapiro, line 280: Did these get removed when we filtered? And/or should we just skip adding them at the earlier step (Line 117)?

analysis/quantifier-comparisons/13-spatial-transcriptomics-benchmarking.Rmd
```diff
@@ -0,0 +1,411 @@
...
+
+
+```{r}
+# total UMI/ spot 
+ggplot(coldata_df_common, aes(x = sum, color = tool)) + 
+  geom_density() + 
+  theme_classic() + 
+  theme(axis.text.x = element_text(angle = 45, hjust = 1)) + 
+  ylab("UMI/spot") + 
+  xlab("")
```
jashapiro, line 245: The UMI count difference here (and genes detected below) seems pretty significant, but I'm not sure what to make of it. (and thank you for using density plots!)

analysis/quantifier-comparisons/13-spatial-transcriptomics-benchmarking.Rmd
```diff
@@ -0,0 +1,411 @@
...
+```{r}
+# mean gene expression across shared genes 
+ggplot(rowdata_cor, aes(x = `spaceranger`, y = `alevin-fry`)) +
+  geom_point(size = 0.5, alpha = 0.1) + 
+  scale_x_log10() + 
+  scale_y_log10() + 
+  labs(x = "Spaceranger mean gene expression", y = "Alevin Fry Mean gene expression") + 
+  theme_classic()
+```
+Correlation appears to be quite high between mean gene expression in Spaceranger and Alevin-fry, however, we do see that there is a subset of genes that have higher gene expression in Spaceranger than in Alevin-fry and are slightly off the diagonal.
```
jashapiro, line 339: Do we have any sense of what these genes are? It seems like a good number, and it is striking how it looks like there are two lines there...

analysis/quantifier-comparisons/13-spatial-transcriptomics-benchmarking.Rmd
```diff
@@ -0,0 +1,411 @@
...
+
+# restrict to only common genes 
+common_genes <- gene_counts %>%
+  dplyr::filter(n == 2) %>%
+  dplyr::pull(gene_id)
+
+rowdata_df_common <- rowdata_df %>%
+  dplyr::filter(
+    (gene_id %in% common_genes) 
+  )
```
jashapiro, line 306: minor formatting
```suggestion
  dplyr::filter(gene_id %in% common_genes) 
```

jashapiro (commented):
This looks good, but I had a few comments in there with suggestions, in particular about the `read10xVisium` function, which I think we might want to avoid. I mostly don't like the idea of having to use a separate function at the end for plotting depending on where the data came from, so I think we should fix the problem at read time.

I think most of my other comments are minor, but it might be that the difference between spaceranger and alevin seems large enough to me to be worthy of comment. I wonder if the final correlation plot might benefit from a x=y line, so that we can see how different the expression levels are on a gene by gene basis? In other words, are the differences in mean expression/UMI count that we see at the spot level driven by an overall difference between the quantifiers for every gene, or is it a smaller subset of genes (the lower stripe) that are driving the differing distributions?

analysis/benchmarking-functions/R/spe_to_df.R
```diff
@@ -0,0 +1,26 @@
...
+    dplyr::mutate(in_tissue = spatialData(spe)$in_tissue)
+}
+
+
+#' Convert rowData from SpatialExperiment object into data.frame
+#'
+#' @param spe SpatialExperiment obejct with per gene QC metrics in rowData
+#'
+#' @return data.frame containing rowData from original spe
+spatial_rowdata_to_df <- function(spe){
```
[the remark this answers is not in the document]
allyhawkins: Yes this is exactly the same and we currently do have functions in `scpcaTools` that grab the `colData` and `rowData` in the package, but they currently only accept a `SingleCellExperiment` and will throw an error if it is not that which is why I added in these functions here temporarily and then we can modify the functions in `scpcaTools` as a separate PR.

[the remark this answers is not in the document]
jashapiro: I just looked at the functions we have, and it seems like they should work? What error are you getting? `SpatialExperiment` is derived from `SingleCellExperiment`, so `is(spe, "SingleCellExperiment")` should return TRUE and the tests should pass.

allyhawkins:
I believe I addressed the majority of your comments that you left @jashapiro. I modified the function that we use to import the spaceranger only files to not use `read10XVisium`, and now do it similar to the way I am importing the fry+ spaceranger files but without subsetting to only spots found in both tools. 

I also removed the function to grab the `rowData` and convert to a data frame and switched to using the function I already had in `scpcaTools`. I did not remove the `colData` function because as part of the new function I wrote here, I am also merging the `colData` with the `spatialData` to get the `in_tissue` column specifically for each spe. I figured in a separate PR if needed I can modify the function in `scpcaTools` to incorporate the `spatialData` or we can just wait until they do the reorganization and then if we need to use these functions again we can use the functions as they are in `scpcaTools`. 

I did notice when going back through and examining things a bit more closely that the Alevin-fry + Spaceranger does tend to have higher counts and genes detected per cell and you can see that in the spatial plots as well that the patterning, although similar, does look denser. I also am not quite sure which one is "correct" or what to think about, but definitely food for thought that these don't match up nearly as nicely as with the single-cell data. 

I also did some playing around with the genes that don't have as high correlations between their mean expression values. After playing around with it a bit, I kept in the calculation of the fold change between mean gene expression in alevin-fry vs. spaceranger and highlighted genes on the graph with high or low fold change. I also did over enrichment analysis and noticed that no significant GO terms came up. When scrolling through the gene names you do see a lot of ribosomal and translation associated genes and then some HOX and developmental genes as well that are on the outer edges of the correlation scatter plot, but I wasn't sure the best way to do this analysis and am open to other ideas. I also played around with a linear regression analysis and plotted the genes with high residuals and those are the same genes that are showing up in this analysis as well. If you had a particular idea of what you were looking for here please let me know. 

[Here is a link to the updated html file for reference.](https://htmlpreview.github.io/?https://github.com/AlexsLemonade/alsf-scpca/blob/allyhawkins/spatial-transcriptomics-benchmarking-2/analysis/quantifier-comparisons/13-spatial-transcriptomics-benchmarking.nb.html)

jashapiro:
I haven't had time for a full review here, but I was looking at the outputs, and I think some error might have been introduced in the latest changes: the images for spaceranger don't look correct now:
Screen Shot 2021-12-02 at 09 46 06
Screen Shot 2021-12-02 at 09 45 56


I feel like the distributions are odd too?
Screen Shot 2021-12-02 at 09 54 23

allyhawkins:
Thanks for catching this @jashapiro! After taking another look at this, it looks like using the `read10XVisium` function is doing something slightly different than how we were creating it manually. I went back through and couldn't quite figure out why that was and it looks like the main difference was that function appears to read in the HDF5 files by default, but I don't think that should be the difference. I removed the `create_spaceranger_spe` function that I had made and am now going back to using `read10XVisium` since that has been fixed and the images now all look as they should. Interestingly, Spaceranger shows higher counts, genes detected, and mean gene expression than Alevin-fry. 

[Here's the updated html](https://htmlpreview.github.io/?https://github.com/AlexsLemonade/alsf-scpca/blob/allyhawkins/spatial-transcriptomics-benchmarking-2/analysis/quantifier-comparisons/13-spatial-transcriptomics-benchmarking.nb.html).

jashapiro:
It looks like you are using the `SpatialExperiment` version from GitHub, which is fine, but I think that should be explicitly noted in the notebook (it is in the `sessioninfo`, but I might not expect people to look there right away). 
(I had a thought finally about the reason for the plots above: the coordinates and cells were presumably not getting read in the same order, so there was some re-sorting that needed to be done. That would make the density plots include spots that were not really in tissue, resulting in the bimodal distribution)

Finally getting to thinking about the results here:
I am a bit surprised by the fact that alevin-fry is performing so much worse. I assume we are using the same settings as with the single-cell data for the mapping and quantification? I looked back at earlier results and it does seem that sometimes there is a slight difference between AF & cellranger, but we haven't seen it to this degree before? 

I know I had asked you to focus on one sample to reduce confusion, but in this case maybe we will need to look at more than one to know how to proceed!

allyhawkins:
> It looks like you are using the SpatialExperiment version from GitHub, which is fine, but I think that should be explicitly noted in the notebook (it is in the sessioninfo, but I might not expect people to look there right away).

I added in a note about this at the beginning of the notebook. 

> I am a bit surprised by the fact that alevin-fry is performing so much worse. I assume we are using the same settings as with the single-cell data for the mapping and quantification? I looked back at earlier results and it does seem that sometimes there is a slight difference between AF & cellranger, but we haven't seen it to this degree before?

> I know I had asked you to focus on one sample to reduce confusion, but in this case maybe we will need to look at more than one to know how to proceed!

I also was surprised by the difference here as with the single-cell and single-nuclei we saw a lot more overlap between cellranger and alevin-fry. I added the second sample back in here and I am still seeing the same trend. I am not so inclined to use Alevin-fry here and am not seeing a particular reason to include it, as Spaceranger seems to be capturing more information.

jashapiro (approved):
Looks good! I had just a couple minor comments about content, and then some additional thoughts about results.

I think it may be worth rerunning this analysis after creating the updated cellranger/spaceranger index & rerunning, just to see what might have changed. I don't expect the main results to change, but it is possible that some of the more curious patterns for a few genes might. Or might not. 

If we do revisit this, it probably makes sense to update to use alevin-fry unfiltered as well, [as discussed earlier](https://github.com/AlexsLemonade/alsf-scpca/pull/146#pullrequestreview-806412324). I don't expect that to make much of a difference, but it could make a difference if some of the knee-inferred barcodes are close enough to the true spot barcodes that they would be combined under that method.

I just had a thought about how we might see if that is a likely problem. If the AF knee method is keeping some spot barcodes that do not correspond to real spots, we should see those in the set of barcodes that are removed when joining with spot coordinates. If there are barcodes that are being removed that have "surprisingly high" UMI counts, that might indicate that this is part of the difference. I don't think this needs to be part of this PR, but I think we might want to have a look before deciding that AF should not be used. (While it seems we do have to run spaceranger no matter what, having AF results for comparison to the single cell data seems like it still might be worthwhile).

analysis/quantifier-comparisons/13-spatial-transcriptomics-benchmarking.Rmd
```diff
@@ -0,0 +1,534 @@
...
+  scale_x_upset(n_intersections = 3)
+```
+
+Similar to with the single-cell data most genes are detected in both Alevin-fry and Spaceranger which is good. 
+
+## Concluding thoughts 
+
+- Alevin-fry and Spaceranger only result in similar distributions of UMI/spot and genes/spot, although it is not quite as nice as the overlay you see with single-cell libraries. 
+- It appears that Spaceranger has slightly higher UMI/cell and genes detected/cell. 
+- Both tools also show high correlation in mean gene expression and high overlap in the genes that are detected, although there is an increase in gene expression in Spaceranger alone.
```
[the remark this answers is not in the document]
allyhawkins: Yea I was trying to figure out a good way to identify those specific genes which is why I had done the analysis of looking at fold change of gene expression between spaceranger and alevin-fry. So if you overcrowd the plots with labels they will get labeled as those that have a change in expression, but it looks horrible and I was struggling to come up with filtering conditions to only label those points where it was informative. I will make a note of the index in this notebook though.

PR #156, opened 2021-12-02
Fix errors in spaceranger Dockerfile

allyhawkins:
In testing the Spaceranger workflow that I was adding in `scpca-nf`, I noticed that the docker image that we had for spaceranger was not working properly and did not have Spaceranger properly installed on it. Here, I added one line to go into the correct working directory before copying over spaceranger so that when it gets added to the home path it is done so correctly. I went into the image and ran a test run of spaceranger and everything now works as expected.


jashapiro (approved):
🚀

PR #160, opened 2021-12-10
Make genetic-demux into a full workflow

jashapiro:
This PR creates a `genetic-demux.nf` workflow that combines the earlier individual steps of genetic demultiplexing into a unified workflow that can run from start to finish with only specifying the input run to be processed.

In doing this, I deleted most of the test run setup from earlier workflows, created named workflows for each step with `take` and `emit` statements, and centralized parameters for specifying containers, files, and directories. So while there are a lot of changes, many of them are moving things around a bit and deleting mock data.

The biggest change/additions are:
1. Adding logic to identify and build a channel of the bulk samples that are needed for reference genotyping. 
2. the logic for grouping samples by multiplex library for `mpileup`
3. updating the workflow for cellsnp/vireo, which now includes more logic to pair the vcf files generated by  `mpileup` with the `starsolo` runs. 

I tried to leave comments to explain what is going on in each of those steps, but I might need to add more. nextflow can be a bit obtuse.

When running this, I found that some tweaks to the overall config were needed, so there are a few changes there as well.

Note that there is one process here that is not used: I was doing some testing with calling initial SNPs from the bulk data with cellsnp as well (what they call mode 2b). This is much faster than `mpileup`, but seemed in initial testing to give worse results, I may do a bit more testing with it, so I left the process there, even though no workflow currently uses it.

### DIFF
--- a/workflows/genetic-demux/cellsnp.nf
+++ b/workflows/genetic-demux/cellsnp.nf
@@ -67,12 +67,13 @@ workflow cellsnp_vireo {
     mpileup_vcf_ch // channel of [meta, vcf_file]
   main:
     mpileup_ch = mpileup_vcf_ch
-      .map{[it[0].multiplex_library_id, it[0], it[1]]} // pull out library id for combining
+      .map{[it[0].multiplex_library_id] + it} // pull out library id for combining
     star_mpileup_ch = starsolo_bam_ch
-      .combine(starsolo_quant_ch, by: 0)
-      .map{[it[0].library_id, it[0], it[1], it[2]]} // add library id at start
-      .combine(mpileup_ch, by: 0)
-      .map{[it[1], it[2], it[3], it[4], it[5], it[6]]} // drop library id
+      .combine(starsolo_quant_ch, by: 0) // rejoin starsolo outs by meta object
+      .map{[it[0].library_id] + it} // add library id at start
+      .combine(mpileup_ch, by: 0) // join starsolo and mpileup by library id
+      .map{it.drop(1)} // drop library id
+      //result: [meta, star_bam, star_bai, star_quant, meta_mpileup, vcf_file]
     
     cellsnp(star_mpileup_ch)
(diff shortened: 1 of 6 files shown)


workflows/genetic-demux/cellsnp.nf
```diff
@@ -112,11 +52,23 @@ process vireo{
...
+      .map{[it[0].multiplex_library_id] + it} // pull out library id for combining
+    star_mpileup_ch = starsolo_bam_ch
+      .combine(starsolo_quant_ch, by: 0) // rejoin starsolo outs by meta object
+      .map{[it[0].library_id] + it} // add library id at start
+      .combine(mpileup_ch, by: 0) // join starsolo and mpileup by library id
+      .map{it.drop(1)} // drop library id
+      //result: [meta, star_bam, star_bai, star_quant, meta_mpileup, vcf_file]
+    
+    cellsnp(star_mpileup_ch)
+    vireo(cellsnp.out)
```
allyhawkins, line 71:
```suggestion
    cellsnp(star_mpileup_ch) \
      | vireo
```
You could use the pipe here if you wanted

workflows/genetic-demux/mpileup.nf
```diff
@@ -42,92 +30,74 @@ process mpileup{
     """
 }
 
+process mpileup_cellsnp{
```
allyhawkins, line 33: If this isn't currently being used right now can we add a comment that states that it's still being tested or not currently used in the workflow?

allyhawkins (approved):
I just had one comment about trying to remove the duplication of the `cell_barcodes` map and then a suggestion for use of the pipe but other than that looks good to go to me! I couldn't suggest all the changes related to `cell_barcodes` because they weren't close enough to changed lines but made a note in my comment.

jashapiro:
Merging as it is working now & I incorporated all feedback. Will probably need future tweaks!

Issue #157, opened 2021-12-07
Rerun spatial transcriptomics benchmarking with new cellranger index


allyhawkins:
From https://github.com/AlexsLemonade/alsf-scpca/pull/151#pullrequestreview-825323348 we may want to rerun the benchmarking notebook, `13-spatial-transcriptomics-benchmarking.Rmd` after generating a new cellranger index using ensembl 104 to see if any of the results change. In particular if any of the effects on particular genes are altered.

If we do revisit this, we may also want to add in Alevin-fry unfiltered rather than using the knee filtering method as well before we discount using Alevin-fry.

PR #161, opened 2021-12-10
Modify alevin-fry to use unfiltered mode with spatial libraries + update default indices 

### LINKED ISSUE #158: Modify alevin-fry workflow to run spatial transcriptomics libraries using unfiltered mode
As discussed in https://github.com/AlexsLemonade/alsf-scpca/pull/151#pullrequestreview-825323348, we should try and use Alevin-fry with the unfiltered mode rather than the knee mode for ST libraries and see if that affects some of the results we are seeing in benchmarking. 

For some more context, this was originally posted by @jashapiro in https://github.com/AlexsLemonade/alsf-scpca/pull/151#pullrequestreview-825323348: 

> I just had a thought about how we might see if that is a likely problem. If the AF knee method is keeping some spot barcodes that do not correspond to real spots, we should see those in the set of barcodes that are removed when joining with spot coordinates. If there are barcodes that are being removed that have "surprisingly high" UMI counts, that might indicate that this is part of the difference. I don't think this needs to be part of this PR, but I think we might want to have a look before deciding that AF should not be used. (While it seems we do have to run spaceranger no matter what, having AF results for comparison to the single cell data seems like it still might be worthwhile).

The first step in doing this should be modifying the current workflow to allow for the unfiltered mode to work with ST libraries in alevin-fry. As a part of this, we will need to grab the barcode files for visium from `spaceranger-1.3.1/lib/python/cellranger/barcodes/` and add them to the barcodes folder on s3, `s3://nextflow-ccdl-data/reference/10X/barcodes`. The barcode files then need to be added to the list of barcode files in the actual workflow. 

Then we need to change the technology in the metadata to reflect the version of visium that was used, according to the spaceranger output for `SCPCR000372` and `SCPCR000373` (the samples being used for benchmarking), the version is `visium_v1`. Changing the technology is important to make sure we are using the correct barcode file.

allyhawkins:
Closes #158 and also partially addresses #157. 

The first thing that I have done here is modify the alevin-fry workflow to work with spatial libraries with the unfiltered mode in Alevin-fry. To do this I added the barcode lists for visium_v1 and visium_v2 to `s3://nextflow-ccdl-data/reference/10X/barcodes`. I then added the corresponding filenames for the barcode files to the barcodes map and included the `visium_v1` and `visium_v2` options in the technology map. This was all that was needed to allow for unfiltered to work and I tested it with the two benchmarking samples and everything ran smoothly. 

While modifying the alevin-fry workflow, I noticed that it was still using the ensembl version 103. Because as part of #157, we want to make sure we are comparing libraries that have been mapped with indices created from the same ensembl version, I updated both the alevin-fry and cellranger/spaceranger workflows to use the ensembl 104 as the default index (we don't have a cellranger ensembl 103 index). Because the name of the cellranger index includes `.`, it will throw an error in spaceranger since the `--id` option can only include letters, digits, underscores, and dashes. To get around this, I added an `index_name` parameter that is a shortened version of the index directory that can be used to tag the output folder when used during benchmarking. I also combined all of the individual index parameters into one `index_path`.

As a result of this PR we can now include Alevin-fry knee, Alevin-fry unfiltered, and Spaceranger all mapped to the same ensembl version in the ST benchmarking notebook as the next steps.

### DIFF
--- a/workflows/alevin-quant/run-alevin-fry.nf
+++ b/workflows/alevin-quant/run-alevin-fry.nf
@@ -37,7 +37,7 @@ barcodes = ['10Xv2': '737K-august-2016.txt',
             '10Xv3.1': '3M-february-2018.txt',
             '10Xv2_5prime': '737K-august-2016.txt',
             'visium_v1': 'visium-v1.txt',
-            'visium_v2': 'visium-v1.txt',
+            'visium_v2': 'visium-v2.txt',
             'spatial': '']
 
 // supported single cell technologies
--- a/workflows/alevin-quant/run-alevin-fry.nf
+++ b/workflows/alevin-quant/run-alevin-fry.nf
@@ -2,13 +2,13 @@
 nextflow.enable.dsl=2
 
 // run parameters
-params.ref_dir = 's3://nextflow-ccdl-data/reference/homo_sapiens/ensembl-103'
+params.ref_dir = 's3://nextflow-ccdl-data/reference/homo_sapiens/ensembl-104'
 params.index_dir = 'salmon_index'
 params.annotation_dir = 'annotation'
 params.index_type = 'cdna' // default index type is cdna, can also use splici
 params.sketch = false // use sketch mode for mapping with flag `--sketch`
 params.resolution = 'full' //default resolution is full, can also use cr-like, cr-like-em, parsimony, and trivial
-params.t2g_3col = 'Homo_sapiens.GRCh38.103.spliced_intron.tx2gene_3col.tsv'
+params.t2g_3col = 'Homo_sapiens.GRCh38.104.spliced_intron.tx2gene_3col.tsv'
 params.barcode_dir = 's3://nextflow-ccdl-data/reference/10X/barcodes' 
 params.filter = 'unfiltered' // default filtering strategy is to use unfiltered, other options include knee (--knee-distance)
 params.run_metafile = 's3://ccdl-scpca-data/sample_info/scpca-library-metadata.tsv'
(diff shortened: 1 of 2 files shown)


workflows/alevin-quant/run-alevin-fry.nf
```diff
@@ -24,18 +24,20 @@ ALEVINFRY_CONTAINER = 'quay.io/biocontainers/alevin-fry:0.4.1--h7d875b9_0'
...
+t2g_map = ['cdna': 'Homo_sapiens.GRCh38.104.spliced.tx2gene.tsv',
+           'splici': 'Homo_sapiens.GRCh38.104.spliced_intron.tx2gene.tsv']
 
 // 10X barcode files
 barcodes = ['10Xv2': '737K-august-2016.txt',
             '10Xv3': '3M-february-2018.txt',
             '10Xv3.1': '3M-february-2018.txt',
             '10Xv2_5prime': '737K-august-2016.txt',
+            'visium_v1': 'visium-v1.txt',
+            'visium_v2': 'visium-v1.txt',
```
jashapiro, line 40: I assume? But fun fact that I just discovered... the two files are identical.
```suggestion
            'visium_v2': 'visium-v2.txt',
```

workflows/cellranger-quant/run-cellranger.nf
```diff
@@ -17,22 +16,20 @@ cellranger_tech_list = ["10Xv2", "10Xv3", "10Xv3.1", "10Xv2_5prime"]
...
   container '589864003899.dkr.ecr.us-east-1.amazonaws.com/scpca-cellranger:6.1.2'
   publishDir "${params.outdir}", mode: 'copy'
-  tag "${meta.scpca_run_id}-${index}" // add tag for tracking sample names in trace file  
+  tag "${meta.scpca_run_id}-${index_name}" // add tag for tracking sample names in trace file  
   label 'cpus_8'
   label 'bigdisk'
   input:
     tuple val(meta), path(fastq_dir), val(include_introns)
+    val index_name
     path index
```
jashapiro, line 28: I'd probably make this a tuple for organizational logic purposes, though I don't think it matters much.
```suggestion
    tuple val(index_name), path(index)
```

workflows/cellranger-quant/run-cellranger.nf
```diff
@@ -120,9 +118,9 @@ workflow{
                        )}
 
   // run cellranger
-  cellranger(cellranger_reads, params.index_path)
+  cellranger(cellranger_reads, params.index_name, params.index_path)
```
jashapiro, line 121: If you follow my tuple suggestion at line 27, then this becomes: 
```suggestion
  cellranger(cellranger_reads, [params.index_name, params.index_path])
```

workflows/cellranger-quant/run-cellranger.nf
```diff
@@ -49,16 +46,17 @@ process cellranger{
...
   container '589864003899.dkr.ecr.us-east-1.amazonaws.com/scpca-spaceranger:1.3.1'
   publishDir "${params.outdir}", mode: 'copy'
-  tag "${meta.scpca_run_id}-${index}-spatial" 
+  tag "${meta.scpca_run_id}-${index_name}-spatial" 
   label 'cpus_8'
   label 'bigdisk'
   input:
     tuple val(meta), path(fastq_dir), file(image_file)
+    val index_name
     path index
```
jashapiro, line 55: Repeating the same suggestion about grouping index_name and index
```suggestion
    tuple val(index_name), path(index)
```
```

File: workflows/cellranger-quant/run-cellranger.nf [hunk 5]
```diff
@@ -120,9 +118,9 @@ workflow{
                        )}
 
   // run cellranger
-  cellranger(cellranger_reads, params.index_path)
+  cellranger(cellranger_reads, params.index_name, params.index_path)
 
   // run spaceranger 
-  spaceranger(spaceranger_reads, params.index_path)
+  spaceranger(spaceranger_reads, params.index_name, params.index_path)
```
jashapiro, line 124: Finishing repeat of suggestion
```suggestion
  spaceranger(spaceranger_reads, [params.index_name, params.index_path])
```

jashapiro (approved):
LGTM!

I just made a couple minor suggestions: A "correction" about the barcode file name, and an organizational suggestion (making `index_name` and `index` a tuple) that applies both to the `cellranger` and `spaceranger` processes, if you decide to use it. I was going to only suggest in one place, but I figured it is easier to change all at once in github if you should choose to do so.

PR #162, opened 2021-12-13
Update Spatial transcriptomics benchmarking with Alevin-fry unfiltered 

### LINKED ISSUE #159: Include alevin-fry unfiltered for Spatial in ST benchmarking
After completing #158 we should run the same two benchmarking samples, SCPCR000372 and SCPCR000373, through the Alevin-fry workflow using the unfiltered mode and add them to our benchmarking results. Here we will want to make sure that we perform filtering on the unfiltered output before performing any comparisons. 

When reading in alevin-fry results, only those spots that are also identified with Spaceranger will be included. The next filtering should involve removing any spots that do not include tissue. This can be done by removing any spots where `spatialData$in_tissue == 0`. Based on the [documentation for Spaceranger](https://support.10xgenomics.com/spatial-gene-expression/software/pipelines/latest/output/matrices), it appears that the only filtering they do between the raw and filtered output is by removing spots that don't have tissue overlapping so we should not need any additional filtering beyond that. 

Comparisons should be made between Alevin-fry knee, Alevin-fry unfiltered, and Spaceranger by modifying the current benchmarking notebook, `analysis/quantifier-comparisons/13-spatial-transcriptomics-benchmarking.Rmd`.

allyhawkins:
Closes #159 and #157. This PR addresses two remaining questions for benchmarking of spatial samples: 

1. The same samples we have been using for benchmarking, SCPCR000372 and SCPCR000373, were quantified using Alevin-fry with an unfiltered permit list of all possible spot barcodes and then added to the previous comparisons of Alevin-fry knee filtering and Spaceranger. 
2. All workflows have been updated to be using an index that was generated from the same reference files so have the same possible gene list. All samples that were not already using an index generated using ensembl v104 were re-run so now all samples can be more easily compared. 

After re-running the samples accordingly, I updated the analysis notebook, `13-spatial-transcriptomics-benchmarking.Rmd` to display results from Alevin-fry-knee, Alevin-fry-unfiltered, and Spaceranger for both benchmarking samples. In addition to making the changes needed to show the plots with all three tools (rather than just the two included previously), I made some other additions to the anlaysis: 

1. I narrowed in on the group of genes that are appear to be "off the diagonal" in the correlation plots, plotting them by labeling them with two different colors. This allowed me to identify the criteria that specifically categorized that group of genes and then I pulled that group of genes to perform over-representation analysis. I noted that there were no significant gene lists that were enriched in that specific group of genes, with no particular pathway or gene list showing increased gene expression in Spaceranger over either of the Alevin-fry methods. 
2. I also looked at the group of genes that is uniquely quantified in Spaceranger, but not found in Alevin-fry and found again no specific pathways or gene lists that were enriched. However, when looking at the actual gene names it appears that the majority of them are ribosomal genes, mitochondrial genes, or long intergenic noncoding genes, so nothing that we would particularly be interested in keeping. 

** Note that previously I had separated the ORA by sample, but I thought it made sense to only include genes that were found to have altered expression between the two tools in both samples, eliminating the two pathways that had previously been enriched in SCPCR000372 only. Let me know if you would prefer me to look at each sample individually instead. 

After adding in Alevin-fry-unfiltered, I noticed that Alevin-fry was able to detect all the same spots as Spaceranger, with no loss of spots, while Alevin-fry-knee appears to lose some spots that should be there. The actual distribution of UMI/cell and genes detected/cell in alevin-fry-unfiltered appears to not be affected by this and looks to be almost identical to alevin-fry-knee. 
I'm struggling with another reason behind why Alevin-fry is resulting in decreased counts and genes/ cell than Spaceranger. I would expect that by using the `cr-like-em` resolution that we are using that we would actually be resolving some multi-mapped reads that spaceranger may not be resolving, which would result in potentially higher number of UMIs not less. 

[Here is the updated html of the analysis.](https://htmlpreview.github.io/?https://github.com/AlexsLemonade/alsf-scpca/blob/allyhawkins/benchmarking-spatial-fry-unfiltered/analysis/quantifier-comparisons/13-spatial-transcriptomics-benchmarking.nb.html)

### DIFF
--- a/analysis/quantifier-comparisons/13-spatial-transcriptomics-benchmarking.Rmd
+++ b/analysis/quantifier-comparisons/13-spatial-transcriptomics-benchmarking.Rmd
@@ -58,7 +58,7 @@ sample_ids <- c("SCPCR000372", "SCPCR000373")
 
 
 ```{r}
-mito_file <- file.path(base_dir, "sample-info", "Homo_sapiens.GRCh38.103.mitogenes.txt")
+mito_file <- file.path(base_dir, "sample-info", "Homo_sapiens.GRCh38.104.mitogenes.txt")
   
 # read in mito genes 
 mito_genes <- readr::read_tsv(mito_file, col_names = "gene_id")
@@ -503,7 +503,7 @@ background_genes <- rowdata_cor$symbol %>%
 ```
 
 ```{r}
-knee_go_ora_results <- enrichGO(gene = knee_outlier_genes,
+knee_go_ora_results <- enrichGO(gene = knee_diff_genes,
                                 universe = background_genes,
                                 keyType = "SYMBOL",
                                 OrgDb = org.Hs.eg.db,
@@ -511,7 +511,7 @@ knee_go_ora_results <- enrichGO(gene = knee_outlier_genes,
                                 pAdjustMethod = "BH",
                                 pvalueCutoff = 0.00001)
 
-unfiltered_go_ora_results <- enrichGO(gene = unfiltered_outlier_genes,
+unfiltered_go_ora_results <- enrichGO(gene = unfiltered_diff_genes,
                                       universe = background_genes,
                                       keyType = "SYMBOL",
                                       OrgDb = org.Hs.eg.db,
@@ -656,18 +656,43 @@ ggplot(gene_detect_df, aes(x = tools_detected)) +
 
 For the most part genes are found in all tools, however it looks like there is a chunk of genes that we would miss that are found in Spaceranger that are not found in either of the Alevin-fry tools. 
 Are any of those genes involved in important pathways that we might want to make sure that we don't lose? 
+First, we can see what types of genes there are by grabbing the gene annotation from the gtf file. 
+
+```{r}
+# load gtf file and select gene name and gene_biotype
+gtf_file <- file.path(base_dir, "sample-info", "Homo_sapiens.GRCh38.104.gtf.gz")
+gtf <- rtracklayer::import(gtf_file, feature.type = "gene") %>%
+  as.data.frame() %>%
+  dplyr::select(gene_id, gene_name, gene_biotype)
+```
+
+```{r}
+# join spaceranger only genes with gene biotypes from gtf
+spaceranger_only_gene_table <- gene_detect_df %>%
+  dplyr::filter(tools_detected == "spaceranger") %>%
+  dplyr::select(gene_id) %>%
+  dplyr::left_join(gtf, by= "gene_id") 
+
+# create a table of gene biotypes that are specific to spaceranger
+spaceranger_gene_counts <- spaceranger_only_gene_table %>%
+  dplyr::count(gene_biotype) %>% 
+  # reorder by gene biotype
+  dplyr::mutate(gene_biotype =factor(gene_biotype, levels = gene_biotype[order(n)]))
+```
+
+```{r}
+ggplot(spaceranger_gene_counts, aes(y = gene_biotype, x = n)) + 
+  geom_bar(stat = "identity") + 
+  xlab("")
+```
+It looks like the majority of the genes that we lose from not using Spaceranger correspond to pseudogenes and long-noncoding RNA's, with protein coding genes being the third most represented category. 
+We can also do over-representation analysis and see if any genes in specific pathways are lost. 
 
 ```{r}
 # spaceranger only genes to do ORA 
-spaceranger_only_genes <- gene_detect_df %>%
-  dplyr::filter(tools_detected == "spaceranger") %>%
-  dplyr::select(gene_id) %>%
-  dplyr::left_join(gene_symbols_df, by= "gene_id") %>%
-  tidyr::drop_na() %>%
+spaceranger_only_genes <- spaceranger_only_gene_table %>%
   unique() %>%
-  dplyr::pull(symbol)
-
-spaceranger_only_genes
+  dplyr::pull(gene_name)
 ```
 
 ```{r}
@@ -689,7 +714,6 @@ spaceranger_go_results <- spaceranger_ora_results@result %>%
 spaceranger_go_results
 ```
 
-The genes that are identified in only Spaceranger are not enriched for any specific pathway and also appear to be comprised of a lot of ribosomal genes, mitochondrial genes, and long intergenic non-coding RNA's. 
 
 ```{r}
 # save spe list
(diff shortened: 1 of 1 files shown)


analysis/quantifier-comparisons/13-spatial-transcriptomics-benchmarking.Rmd
```diff
@@ -504,23 +651,59 @@ gene_detect_df <- rowdata_df_filtered %>%
...
+
+```{r}
+# look at gene ontology results 
+spaceranger_go_results <- spaceranger_ora_results@result %>%
+  as.data.frame() %>%
+  dplyr::filter(p.adjust < 0.2)
+spaceranger_go_results
+```
+
+The genes that are identified in only Spaceranger are not enriched for any specific pathway and also appear to be comprised of a lot of ribosomal genes, mitochondrial genes, and long intergenic non-coding RNA's.
```
jashapiro, line 692: Is this based mostly on looking at the list? Is there a way we could quantify this easily? If we have a table where the gene type is a column, could we summarize it for a more quantitative look?

jashapiro (approved):
This looks great! It really answered a lot of the questions I had from the previous version. 

I'm also a bit mystified by the lower numbers from alevin-fry here. I wonder if the tissue samples result in more degraded RNA, making mapping with AF less effective. This could also explain why some samples do better than others.) I don't think it is worth testing at this stage, but I wonder if this is is place where smaller kmers would make a difference. 

My only substantive comment is whether we can quantify the kinds of genes that are differentially affected (besides GO). I don't really know what we would do with the info at this point, but it might be worth having a record.

allyhawkins:
> I'm also a bit mystified by the lower numbers from alevin-fry here. I wonder if the tissue samples result in more degraded RNA, making mapping with AF less effective. This could also explain why some samples do better than others.) I don't think it is worth testing at this stage, but I wonder if this is is place where smaller kmers would make a difference.

@jashapiro I think this is a really really good point. I've noticed from experience in trying to capture regions of genes far from the 3' end that there is fall off of gene expression in spatial libraries as you move from the 3' end a lot sooner than when doing similar experiments in single-cell. It's definitely very likely that with the tissue, the RNA is not as intact. 

@rob-p, we have been working on comparing quantification of spatial transcriptomics libraries with Alevin-fry to Spaceranger and have noticed that overall there is a decrease in the number of UMIs and genes detected/cell in Alevin-fry than in Spaceranger. We also have seen that there is a shift in mean gene expression in a subset of genes with this subset showing higher mean gene expression in Spaceranger. Here we are using Salmon version 1.5.2 and Alevin-fry version 0.4.1. We are mapping to the splici index and then using the `cr-like-em` resolution. I have tested using both the knee filtering and the unfiltered permit list and overall see very similar results with the only difference being that the unfiltered version is able to identify all spots that are identified in Spaceranger, while some of those spots are lost in the knee filtering. Both Josh and I have been struggling a bit with trying to understand why the quantification of Alevin-fry might be showing lower counts and gene expression than Spaceranger, and were wondering if this is something you had seen before? or if you have any thoughts as to why this might be the case? 

One thought that we had been discussing was potentially due to the increased degradation of RNA that could be occurring in the tissue, so perhaps using smaller kmers could help resolve this. Right now we are using a kmer size of 31. Do you have any thoughts on if the kmer size would be impacting this and if so if it would be worth decreasing the kmer size? 
[Here is a link to the html report with the findings discussed above.](https://htmlpreview.github.io/?https://github.com/AlexsLemonade/alsf-scpca/blob/allyhawkins/benchmarking-spatial-fry-unfiltered/analysis/quantifier-comparisons/13-spatial-transcriptomics-benchmarking.nb.html)
If you have the time, we would appreciate any feedback or thoughts that you have on this matter as you have been very helpful throughout our benchmarking process. Thank you in advance!

rob-p:
Hi @allyhawkins and @jashapiro --- I probably won't have time to do a deep dive this week, as we are in the middle of finals week and I'm teaching an undergraduate class this semester.  However, if you can run a fairly straightforward test, I had a thought that immediately came to mind.

If the issue is mapping-related, e.g. due to a degraded transcriptome, then there will be more reads that alevin refuses to align in selective-alignment mode, whereas CellRanger may be content taking a big soft clip.  The first way I would test this hypothesis is to run alevin-fry in `--sketch` mode.  In this case, the alignments won't be validated under some minimal required score, and so the mappings should be much more robust to degraded transcripts in such a case.  If the UMI count goes up to about what we expect in `--sketch` mode, that is strong evidence for this hypothesis.  We have a feature pending upstream in salmon that is a new and improved implementation of soft-clipping that would eventually allow recovering this in selective-alignment mode, but `--sketch` is the easiest way to quickly check.

The other thing you can look at is, before even running fry, the read mapping rate coming out of the salmon step.  Presumably, if there are alignment quality issues due to degraded alignments, you'll see a lower overall fragment mapping rate in these samples.  This is also good evidence, though not quite as strong as running in `--sketch`, because this will only give you insight into the alignment loss at the fragment level, not how these ultimately translate to distinct UMI counts or how they are distributed over cells.

analysis/quantifier-comparisons/13-spatial-transcriptomics-benchmarking.Rmd
allyhawkins: After some thinking about this, I went ahead and merged the list of genes with the gtf file and obtained the `gene_biotype`. I added a graph to count the type of genes that are lost and sorted them based on frequency, and the majority of genes that are lost come from long non coding RNA's or pseudogenes with protein coding genes making up the third highest group. Let me know what you think about this approach. 
Screen Shot 2021-12-15 at 4 21 33 PM

allyhawkins:
@rob-p thanks so much for the suggestion! I definitely understand that you are busy so no worries that you don't have time to go through in detail. I was able to run the samples through Alevin-fry with `--sketch` as you had suggested and that seems to have done the trick. It looks like the validation scoring definitely was throwing away some alignments. Thanks again for your help as always!

rob-p:
Hi @allyhawkins, 

  Thanks for testing this out!  I'm glad to see that it eliminated the discrepancy in the UMI / detection rate you observed.  Also, thank *you* for these beautiful and detailed comparisons between methods and parameters, etc.  As we (_hopefully_) grow our user-base for alvein-fry, these will be an invaluable resource in recommending best practices for different types of data, and also a great resource for re-running well-thought-out analyses as we make modifications and improvements to the algorithms and implementation!

PR #163, opened 2021-12-15
Testing quantification of spatial libraries using --sketch alignment with Alevin-fry

allyhawkins:
Following up on https://github.com/AlexsLemonade/alsf-scpca/pull/162#issuecomment-993945130, I went ahead and ran the two samples we have been using with `--sketch` mode for alignment, but keeping the `--unfiltered` mode as that retains the most similarity with Spaceranger. I then added these runs to the existing notebook (in order to not repeat a lot of the same code) and am showing the results grouped by tool + configuration. 

The main output that I was interested in here was checking to see if use of `--sketch` would increase the total counts and number of genes detected/cell observed when quantifying with Alevin-fry and that is in fact the case. The distributions look much more similar to Spaceranger and the plots showing the spots overlaying the tissue look almost identical to Spaceranger. I kept the addition of `--sketch` pretty simple, as it looks to mostly duplicate the results seen with Spaceranger. However you do still see a small subset of genes that shows higher gene expression in Spaceranger, like we were seeing in selective alignment, but to a lesser degree. 

This PR is stacked on #162. 

[The updated html with the new plots can be found here.](https://htmlpreview.github.io/?https://github.com/AlexsLemonade/alsf-scpca/blob/allyhawkins/sketch-spatial/analysis/quantifier-comparisons/13-spatial-transcriptomics-benchmarking.nb.html)

### DIFF
--- a/analysis/quantifier-comparisons/13-spatial-transcriptomics-benchmarking.Rmd
+++ b/analysis/quantifier-comparisons/13-spatial-transcriptomics-benchmarking.Rmd
@@ -789,9 +789,9 @@ readr::write_rds(all_spe_list, spe_file)
 - Alevin-fry and Spaceranger only result in similar distributions of UMI/spot and genes/spot, although it is not quite as nice as the overlay you see with single-cell libraries. 
 - Using alevin-fry-unfiltered results in identification of the same spots that are identified in Spaceranger. 
 - It appears that Spaceranger has slightly higher UMI/cell and genes detected/cell. 
-- The increase in UMI/cell and genes detected/cell in Spaceranger disappears with use of `--sketch` alignment with Alevin-fry. 
+- Using Alevin-fry `--sketch` improves agreement with Spaceranger. 
 - Both tools also show high correlation in mean gene expression and high overlap in the genes that are detected, although there is an increase in gene expression in Spaceranger alone. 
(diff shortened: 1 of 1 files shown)


analysis/quantifier-comparisons/13-spatial-transcriptomics-benchmarking.Rmd
```diff
@@ -727,7 +789,9 @@ readr::write_rds(all_spe_list, spe_file)
 - Alevin-fry and Spaceranger only result in similar distributions of UMI/spot and genes/spot, although it is not quite as nice as the overlay you see with single-cell libraries. 
 - Using alevin-fry-unfiltered results in identification of the same spots that are identified in Spaceranger. 
 - It appears that Spaceranger has slightly higher UMI/cell and genes detected/cell. 
+- The increase in UMI/cell and genes detected/cell in Spaceranger disappears with use of `--sketch` alignment with Alevin-fry. 
 - Both tools also show high correlation in mean gene expression and high overlap in the genes that are detected, although there is an increase in gene expression in Spaceranger alone. 
+- The increase in gene expression found in Spaceranger also decreases through use of `--sketch` alignment and more genes that were quantified in Spaceranger and were not quantified in Alevin-fry with selective alignment are now quantified with `--sketch`.
```
jashapiro, line 794: I might quibble with calling it an "increase in Spaceranger", as I would not want to assign direction/blame to one tool over the other. Maybe just say that using sketch mode improves agreement with spaceranger?

jashapiro (approved):
LGTM! I just had one little wording comment.

PR #165, opened 2021-12-21
Add genetic demultiplexing comparison

jashapiro:
This PR adds a notebook for some quick comparisons between cellhash multiplexing and genetic multiplexing. 

The notebook is designed to be run from the `15a-render-demux.R` script, which loops through a set of samples for which we have both cellhash data and genetic demux results from cellsnp/vireo. 

The comments at the moment are minimal, partly because the nature of running the same notebook on multiple samples makes individual commentary difficult.

That said, the overall result is that genetic demux seems to work for many samples, and the results based on comparison to UMAP clusters seem pretty consistent.

Some analysis that might be missing is exploration of the potential relationship between sequencing depth and call quality, or examination of different thresholds for making calls. 

One more technical note is that `Seurat::HTODemux()` fails a lot (not in the example sample, but others), so we might want to take some time to explore why that is, and whether changing thresholds for `DropletUtils::HashedDrops()` improves it at all. (I have little confidence in `Seurat::MULTIseqDemux()` and might eliminate it from this comparison)


jashapiro:
I've updated this a bunch, mostly to simplify it quite a bit, but also to add some commentary about the comparisons that were done. I expect it still will need a bit more polishing, but it would be nice to have another look.

analysis/quantifier-comparisons/15-demux-comparisons.Rmd
```diff
@@ -0,0 +1,290 @@
...
+  summary()
+```
+
+In general, we see that `DropletUtils::hashedDrops` tends to make many fewer calls than `Seurat::HTODemux`. 
+In either case, we have a large excess of `NA` calls, where the cellhash-based demultiplexing algorithms did not find a single best call. 
+
+This may partly be a function of the quality of the cellhash results for this particular library.
+As seen in `14-cellhash-exploration.Rmd`, the cellhash data for these samples seems to have a lot of background signal that may make accurate demultiplexing based on cellhash data alone difficult.
+
+The genetic demultiplexing calls made by `vireo` seem to identify the most cells, but the relative frequencies of the samples do not necessarily agree with the results from `HTODemux`.
```
allyhawkins, line 108: You may have a better understanding of this than I do since I can only see the notebook generated for this single sample, but for SCPCL000535, `vireo` actually doesn't have the most calls. It has similar calls to Seurat, but technically Seurat has the lowest number of `NA`. Is that the case across the board where the total number of `NA` seem to be more similar between `vireo` and Seurat and then `DropletUtils` is lower? If that's the case I think I would change this comment to make the claim that these two methods are more similar in total calls, but the frequencies of the samples don't always agree. Or maybe say that in the majority of cases, ` vireo` has more calls than Seurat if that's the observation you see?

analysis/quantifier-comparisons/15-demux-comparisons.Rmd
```diff
@@ -0,0 +1,290 @@
...
+This could be because different sets of cells are being called by the two methods, or because the actual calls differ, so we can look in a bit more detail.
+
+## Comparing classifications
+
+To compare the calls between methods, we will build some contingency tables among the call sets.
```
allyhawkins, line 113: Can you add a sentence here that just briefly explains the comparisons that you are making below? I think it would be helpful since you have the function separately to just explicitly state that you are making two comparisons across all the methods, the first is looking at whether or not the two methods agree on how they call cells without considering cells that did not get a call or were not successfully demultiplexed for that method, and the second with considering all cells including those that were not successfully demultiplexed.

analysis/quantifier-comparisons/15-demux-comparisons.Rmd
```diff
@@ -0,0 +1,290 @@
...
+
+`vireo`, which bases its calls on the genetic variants detected in each cell shows a stronger correspondence between sample calls and the expression variation, which makes sense if each sample is likely to have cells with distinct expression profiles.
+There are still clusters with cells from multiple samples, which also seems like it should be expected. 
+
+Overall, it remains unclear which if any of the methods is the most reliable, and attempting to answer that question  is beyond the scope of this analysis, and indeed may be beyond the scope of this project as a whole.
+Rather than presenting a single set of calls as definitive and providing demultiplexed libraries for download, it seems like the best course of action is to allow downloading of the multiplexed library with all samples included, along with the calls and statistics generated by each of the demultiplexing methods.
+The task of deciding among the methods will be left to the end user, using the included statistics to allow for adjusting any thresholds and recalling cells as they deem fit.
+For the HTO-based methods, all of the count data will also be present to allow the application other demultiplexing methods.
```
allyhawkins, line 284:
```suggestion
For the HTO-based methods, all of the count data will also be present to allow the application of other demultiplexing methods.
```
Just a small typo.

allyhawkins (commented):
This looks pretty good and I like the overall flow of the analysis a lot. I just had a few minor suggestions below to clarify some of the comments. 

The other question I had was about if you were planning on including the reports for each individual sample anywhere since you set up the script to run all of them through the template?

analysis/quantifier-comparisons/15-demux-comparisons.Rmd
jashapiro: Oh, I think the one I uploaded was 537, actually, but I meant it to be 535. But yeah, I can/should rephrase to be a bit more general. (Also adding a bit of code to write out the library name).

PR #166, opened 2022-04-25
Update scripts to new metadata format

jashapiro:
We changed the format of the metadata file to use the more generic `files`directory` instead of `s3_prefix`, so I am here updating the workflows to handle that change. 

This also required a change to the check_md5_s3.py script, which is included here as well!


allyhawkins (approved):
LGTM 🚀

PR #168, opened 2022-05-17
Add autoscaling volumes for batch.

arkid15r:
Configure new batch queue.
Create a separate compute enviromnent.
Add launch template with pre-configured user_data.
Add cloud-init config to download and install the auto-scaling daemon.
Configure related IAM policies/roles.
Configure volume remounting to /home.


aws/nextflow-launch-templates.tf
```diff
@@ -29,3 +29,20 @@ resource "aws_launch_template" "nf_lt_bigdisk" {
...
+
+resource "aws_launch_template" "nf_lt_auto_scaled_ebs" {
+  name = "nextflow-launchtemplate-auto-scaled-ebs"
+  tags = var.default_tags
+  # Amazon Linux 2 Kernel 5.10 AMI 2.0.20220426.0 x86_64 HVM gp2
+  image_id = "ami-0022f774911c1d690"
+  block_device_mappings {
+    device_name = "/dev/xvda"
+    ebs {
+      volume_size           = 10  # GiB
```
jashapiro, line 41: Do we want to specify gp3?
```suggestion
      volume_size           = 10  # GiB
      volume_type          = "gp3"
```

aws/nextflow-launch-templates.tf
```diff
@@ -29,3 +29,20 @@ resource "aws_launch_template" "nf_lt_bigdisk" {
   }
   update_default_version = true
 }
+
+resource "aws_launch_template" "nf_lt_auto_scaled_ebs" {
+  name = "nextflow-launchtemplate-auto-scaled-ebs"
+  tags = var.default_tags
+  # Amazon Linux 2 Kernel 5.10 AMI 2.0.20220426.0 x86_64 HVM gp2
+  image_id = "ami-0022f774911c1d690"
```
jashapiro, line 37: Based on the instructions [here](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/ecs-optimized_AMI.html#ecs-optimized-ami-linux-variants.html) I ran `aws ssm get-parameters --names /aws/service/ecs/optimized-ami/amazon-linux-2/recommended` and got a different ami: `ami-061c10a2cb32f3491`. I don't know if this is maybe a region difference?

jashapiro (commented):
I left some small comments below, but when I went to test it in the current form, no jobs started, and looking at the [AWS web interface](https://us-east-1.console.aws.amazon.com/batch/home?region=us-east-1#queues/detail/arn:aws:batch:us-east-1:589864003899:job-queue/nextflow-batch-auto-scaled-ebs-queue) I saw the following error: `CLIENT_ERROR - Launch Template UserData is not MIME multipart format`, which I assume is something about how the `user_data` is set?

aws/nextflow-launch-templates.tf
arkid15r: Specified gp3.

aws/nextflow-batch.tf
```diff
@@ -36,6 +36,16 @@ resource "aws_batch_job_queue" "nf_bigdisk_queue" {
   ]
 }
 
+resource "aws_batch_job_queue" "nf_auto_scaled_ebs_queue" {
+  name = "nextflow-batch-auto-scaled-ebs-queue"
```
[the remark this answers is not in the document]
arkid15r: Renamed.

aws/nextflow-launch-templates.tf
```diff
@@ -29,3 +29,20 @@ resource "aws_launch_template" "nf_lt_bigdisk" {
   }
   update_default_version = true
 }
+
+resource "aws_launch_template" "nf_lt_auto_scaled_ebs" {
+  name = "nextflow-launchtemplate-auto-scaled-ebs"
+  tags = var.default_tags
+  image_id = "ami-01c08d4de548477df"  # ccdl-nextflow-base-v2.0 image
```
jashapiro, line 36: I see that you changed this back, which may be fine, though I worry it might be the cause of the issues I encountered. I realized that I needed to change the location of the aws cli in nextflow settings, which I had neglected to do, and might be having some effect, though I don't think this is the main problem. I worry some about having two version of the aws cli, one of which is presumably older and may be missing features?

https://github.com/AlexsLemonade/scpca-nf/blob/ef585fd1ab13ef24920a39579e5fb61c2c7a263c/config/profile_awsbatch.config#L5

One note is that according to the nextflow docs, the aws cli is mounted into the docker image in kind of a strange way, which could also cause trouble: https://www.nextflow.io/docs/latest/aws.html#aws-cli

aws/user_data/auto_scaled_ebs
```diff
@@ -0,0 +1,30 @@
...
+  - jq
+  - lvm2
+  - rsync
+  - sed
+  - unzip
+  - wget
+
+runcmd:
+  - curl -s "https://awscli.amazonaws.com/awscli-exe-linux-x86_64.zip" -o "/tmp/awscliv2.zip"
+  - unzip -q /tmp/awscliv2.zip -d /tmp && /tmp/aws/install
```
jashapiro, line 23: Looking at the install location, it appears that the binary gets installed in `/usr/local/aws-cli` with symlinks to `/usr/local/bin`. Because of the way [nextflow mounts the aws directory]( https://www.nextflow.io/docs/latest/aws.html#aws-cli ), we might be able to get away with making a link from `/usr/local/aws-cli/aws` to `/usr/local/aws-cli/bin/aws`. This will result in nextflow mounting all of `/usr/local/aws-cli` which should not clobber anything. (If we were to specify the path to nextflow as `/usr/local/bin/aws` it would try to mount all of `/usr/local` in the docker images, which would probably be bad).

jashapiro, line 23: Looking at the install location, it appears that the binary gets installed in `/usr/local/aws-cli` with symlinks to `/usr/local/bin`. Because of the way [nextflow mounts the aws directory]( https://www.nextflow.io/docs/latest/aws.html#aws-cli ), we might be able to get away with making a link from `/usr/local/aws-cli/aws` to `/usr/local/aws-cli/bin/aws`. This will result in nextflow mounting all of `/usr/local/aws-cli` which should not clobber anything. (If we were to specify the path to nextflow as `/usr/local/bin/aws` it would try to mount all of `/usr/local` in the docker images, which would probably be bad).

jashapiro (commented):
I tested this with a small job and the overall setup works as far as submitting jobs goes, but unfortunately it does not seem to be doing the disk scaling as expected when I try to run a larger job.

The command I used to test the larger jobs was:
```
nextflow run main.nf -profile batch,ccdl -with-tower --repeat_mapping --run_ids SCPCR000243,SCPCR000244,SCPCR000245,SCPCR000246
```

These samples have larger input files, and I see the following errors:   
```
Error executing process > 'bulk_quant_rna:fastp (SCPCL000244-bulk)'

Caused by:
  Task failed to start - CannotCreateContainerError: Error response from daemon: mkdir /var/lib/docker/overlay2/ccdcbd5bba38bf81d9ae59ff99ac337db15078e441676c2adefc3faca23a491b-init: no space left on device
```

```
nxf-scratch-dir ip-10-1-1-139.ec2.internal:/tmp/nxf.XXXXrZGOT7
  download failed: s3://ccdl-scpca-data/runs/SCPCR000244/SJWLM051023_D2_S2_L001_R1_001.fastq.gz to ./SJWLM051023_D2_S2_L001_R1_001.fastq.gz [Errno 28] No space left on device
```

I think the key part there is that nextflow is putting the data files in `/tmp`, so that is where the autoexpanding volume should be mounted. That may simplify things a bit (no copying of `/home`?) but I do have a small concern that other directories may still get files. (I think the docker image is a result, not the main cause, as the docker images should be relatively small, but I can't be sure.)

I will also note that I didn't see any evidence on the aws console of new volumes being created, which makes sense if the files were ending up in a different directory that was not part of the expandable mount.

I also made some notes about the aws-cli location, which might have caused some earlier trouble, but could probably be addressed in a separate issue, as I think you were suggesting.

aws/nextflow-launch-templates.tf
arkid15r: Keeping the old one so far. Filed https://github.com/AlexsLemonade/alsf-scpca/issues/169

aws/user_data/auto_scaled_ebs
arkid15r: Copied over to the https://github.com/AlexsLemonade/alsf-scpca/issues/169

jashapiro (commented):
A couple of minor edits to the docs, but with those we should be good to go on this!

PR #170, opened 2022-10-06
Add Vireo image

jashapiro:
In preparation for running workflows without internet access, we need a fully enclosed Vireo image, which did not exist. This PR creates one that is ready for use by Nextflow. 

I tried to keep the package as small as I could, using a slim python base, rather than the conda base I had used previously, but that ended up meaning I needed to install `ps` as well as nextflow needed it.

The README is brief, but does include the fact that the image should be built as an `amd64`, which is not the default on an M1 mac. I decided for this use case not to get into multi-platform builds.


allyhawkins (approved):
This looks good to me and I was able to test building things successfully on my end.

PR #171, opened 2022-11-18
Add script to update checkpoint directories for scpca-nf

jashapiro:
This PR, currently in draft form, is designed to complete https://github.com/AlexsLemonade/ScPCA-admin/issues/408. To do so, the script here performs two tasks: moving checkpoint files from an old to new location (as we changed from `internal` to `checkpoints` directories default) and adding `scpca-meta.json` files to the checkpoint directories as needed. 

In the current form, it processes scRNAseq samples (rad files) and vireo results (which already have scpca-meta.json files in all versions).

The basic idea is to generate the checkpoint directories in the same way scpca-nf (as of the future version 0.4, as I am calling it), then copy file contents from the previous location to the current one. I used the `aws` command line for this part, as `boto3` doesn't have a built-in recursive copy or sync (and I didn't trust myself to write one), though I use boto3 for more atomic operations like checking if files exist and writing the json file.

For files that require the `scpca-meta.json` file, we generate all of the fields that would be created by the workflow, then write that file to the appropriate checkpoint directory.

Before doing either, it checks if there are files in the new location, and will not overwrite unless explicitly told to do so by an option.

Speaking of options, there are many, because there are a lot of things that are relatively constant, but could potentially change. I ended up putting a lot of these into options partly just so I could pass them around as part of the `args` dictionary. 

One thing I don't do at the moment is really check which version of the workflow/nextflow was used at each stage, so everything at the moment gets the . We don't really have that in a file that can be extracted within the checkpoint directory itself, but I could look at the `publish` directory and pull it from the output json file. This is probably worth doing, and will be the next thing I work on.

After that, I will return to the bulk and spatial functions, which exist now mostly as stubs.

Note also that the current `prefix` argument sends us to our sort of working directory. When run for real, this will be changed to `scpca-prod`.

Please do let me know how you think this looks as a general direction, and if you see any other issues or places where the function of the code is unclear.


jashapiro:
> One thing I don't do at the moment is really check which version of the workflow/nextflow was used at each stage, so everything at the moment gets the . We don't really have that in a file that can be extracted within the checkpoint directory itself, but I could look at the `publish` directory and pull it from the output json file. This is probably worth doing, and will be the next thing I work on.

Update: this is now present for the scRNAseq data. It turns out the only thing we seem to have in the output is the workflow version (not nextflow), so that is all I am grabbing.

scripts/update_scpca_checkpoints.py
```diff
@@ -0,0 +1,479 @@
...
+    s3_bucket = s3.Bucket(bucket)
+
+    ### read output metadata, if present
+    try:
+        results_obj = s3_bucket.Object(results_json)
+        results_meta = json.loads(results_obj.get()['Body'].read().decode('utf-8'))
+    except s3.meta.client.exceptions.NoSuchKey:
+        # object did not exist
+        results_meta = {}
+    scpca_version = results_meta.get("workflow_version")
```
allyhawkins, line 62: Will this work if `results_meta` is empty?

scripts/update_scpca_checkpoints.py
```diff
@@ -0,0 +1,479 @@
...
+            "aws", "s3", "sync",
+            f"s3://{bucket}/{origin_prefix}",
+            rad_uri
+        ]
+        subprocess.run(sync_command)
+        ### write json object
+        s3_bucket.put_object(
+            Key = metadata_key,
+            Body = json.dumps(metadata, indent=2)
+        )
```
allyhawkins, line 119: I feel like you could create one function that does builds the metadata and then does the copying. After building the uris and prefixes that are unique to each technology you could have one function that takes as input those uris and the `run` object. That function would create the metadata and do the copying so you wouldn't have the same code repeated.

allyhawkins (commented):
Generally I think this looks good, and I don't have a whole lot of comments on it. My main comment is on refactoring some of it to pull out making the metadata object and copying the files into its own function that you can then run inside each of the individual functions for processing different types of technology.

scripts/update_scpca_checkpoints.py
jashapiro: Yes, it will. `results_meta["workflow_version"]` won't, but `.get()` will return `Null` for a missing key, which is fine.

scripts/update_scpca_checkpoints.py
jashapiro: Yes, I definitely thought about that, but the metadata we need varies slightly among the data types. So I could do _most_ of the values with a function, but then I would need to add the specific values somehow... This way also had the advantage that I could check that I got everything by comparing to an existing file for that checkpoint type relatively easily.

With the writing of files, there were some kind of subtle things with reusing the s3 bucket object that I was trying to avoid.

All of these are surmountable, but I'm really hoping that for a script that will  hopefully be run only once, it won't matter. (subtext: I don't want to work on this more than I need to!)

allyhawkins: I now see the `rad_publish_dir` at the bottom of the metadata file that changes for each method, which makes sense why you did it individually.

scripts/update_scpca_checkpoints.py
```diff
@@ -0,0 +1,479 @@
...
+    )
+    args = parser.parse_args()
+
+    ### Read in library file
+    library_df = pandas.read_csv(args.library_file, sep = "\t",  dtype = 'string', keep_default_na=False)
+
+    # add barcode files
+    library_df['barcode_file'] = library_df['technology'].apply(get_barcode, barcode_dir = args.barcode_dir)
+
+    library_df = library_df.iloc[371:373] # limit to 2 for testing
```
allyhawkins, line 445: Just noting that this should be removed before merging/ running it through for real.

allyhawkins (approved):
Thanks for answering my initial questions, going to go ahead and approve and just left a reminder to remove the line where you subset for testing.

jashapiro:
This script is now done, tested and run!

PR #174, opened 2023-07-10
Script to add ref_mito and ref_fasta_index to scpca-meta.json

### LINKED ISSUE #173: Add `mito_file` and `ref_fasta_index` to scpca checkpoint script
In updating `scpca-nf` to be compatible with running samples from multiple organisms at once, we added new entries to the `scpca-meta.json` file that is produced from the workflow. Because we output and read in that file in different processes, and want to be able to skip mapping if ever re-processing projects that have already been mapped, we need that file to be up to date with the current workflow. 

We should update the `update_scpca_checkpoints.py` script to add those two new entries and ensure that the rest of the contents of the metadata file are the same. Then we will need to run that script to update the existing `scpca-meta.json` files.

allyhawkins:
Closes #173 

Before reprocessing any projects, we need to make sure the `scpca-meta.json` files from previous versions of `scpca-nf` match the `scpca-meta.json` file needed for the latest version (v0.5.2). The difference between these two versions is the addition of `ref_mito` and `ref_fasta_index` to the json files. 

This script runs through all of the libraries in the existing metadata file and does the following:
- Grabs the corresponding `scpca-meta.json` file, using the technology for that run ID to define the path to where the `scpca-meta.json` file should exist. If the file does not exist (as in that sample hasn't been processed), then that sample is skipped. 
- If `ref_mito` and/or `ref_fasta_index` do not exist in the metadata file, those are added to the dictionary holding the json contents. 
- The updated json is uploaded to the same location on S3, overwriting the original file. 

I used the arguments found in `update_scpca_checkpoints.py` for inspiration. Because of that, I have an argument for the `checkpoints_prefix`, which is where the files live that should be modified. I have been testing using `scpca/processed/checkpoints`, but when we run this for real, we will want to use `scpca_prod/checkpoints`.

### DIFF
--- a/scripts/add-refs-scpca-meta.py
+++ b/scripts/add-refs-scpca-meta.py
@@ -1,0 +1,88 @@
+#!/usr/bin/env python3
+
+import argparse
+import json
+import boto3
+import pandas
+
+# Parse command line arguments
+parser = argparse.ArgumentParser()
+parser.add_argument(
+    '--library_file',
+    default='s3://ccdl-scpca-data/sample_info/scpca-library-metadata.tsv',
+    help='path or URI to library data file TSV'
+)
+parser.add_argument(
+    '--bucket',
+    default='nextflow-ccdl-results',
+    help='S3 bucket where results files are located'
+)
+parser.add_argument(
+    '--checkpoints_prefix',
+    default='scpca/processed/checkpoints',
+    help='directory containing checkpoint files to modify'
+)
+parser.add_argument(
+        '--overwrite',
+        action="store_true",
+        help="overwrite existing files at destination"
+    )
+args = parser.parse_args()
+
+# Read in library file
+library_df = pandas.read_csv(args.library_file, sep="\t", keep_default_na=False)
+
+# remove any extra / at the end
+bucket = args.bucket.strip('/')
+checkpoints_prefix = args.checkpoints_prefix.strip('/')
+
+# define techs for setting up checkpionts directories for different modalities
+sc_techs = ["10Xv2", "10Xv2_5prime", "10Xv3", "10Xv3.1"]
+bulk_techs = ['single_end', 'paired_end']
+spatial_techs = ['visium']
+demux_techs = ['cellhash_10Xv2', 'cellhash_10Xv3', 'cellhash_10Xv3.1']
+
+# go through every run id and modify scpca-meta.json file if present
+for run in library_df.itertuples():
+
+    print(f"Processing {run.scpca_run_id}")
+
+    # depending on the technology for that run, define the directory where scpca-meta.json is located
+    if run.technology in sc_techs:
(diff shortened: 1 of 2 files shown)


scripts/add-refs-scpca-meta.py
```diff
@@ -0,0 +1,83 @@
+#!/usr/bin/env python3
+
```
sjspielman, line 2: Could use a quick comment here, something like...

```suggestion
"""
If missing, add `ref_mito` and `ref_fasta_index` fields to `scpca-meta.json` checkpoint files in S3.
"""
```

sjspielman (commented):
I went through the script, including running through (but without writing anything back out!), and it all seems good to me. The main question about the code I have is whether we need to this for CITE technologies as well - I'm pretty sure I encountered previously CITE datasets not running through because those JSON fields were missing, but I don't remember if that was b/c the JSON fields were missing in CITE's json or the corresponding RNA's json.. 

Besides code, the main overall comment I have is docs. We should probably get a `scripts/README` going at this point to briefly (1-2 bullets) explain each script here and the context for which each was written.

scripts/add-refs-scpca-meta.py
sjspielman: Woops, back for one more comment I forgot! We should also include some usage info up here, in particular for providing non-default args. Thinking of this comment you made in the PR - 

> I have been testing using `scpca/processed/checkpoints`, but when we run this for real, we will want to use `scpca_prod/checkpoints`

scripts/add-refs-scpca-meta.py
```diff
@@ -0,0 +1,83 @@
...
+args = parser.parse_args()
+
+# Read in library file
+library_df = pandas.read_csv(args.library_file, sep="\t", keep_default_na=False)
+
+# remove any extra / at the end
+bucket = args.bucket.strip('/')
+checkpoints_prefix = args.checkpoints_prefix.strip('/')
+
+# define techs for setting up checkpionts directories for different modalities
```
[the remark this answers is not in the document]
allyhawkins: In any version pre 0.5.2 we actually didn't output a `-features` checkpoint for mapping. We added it in recently, so when CITE is re-processed it will start with mapping but that processing cost is low.

allyhawkins:
Thanks for the comments @sjspielman. I added some information about usage at the top of the script and a readme to explain how we use/used both of these scripts.

sjspielman (approved):
Just one "typo" to fix and all set!

Looking to next steps, let me know when you've run this through (if not already?) and I can get started on re-processing CITEseq projects.

allyhawkins:
> Looking to next steps, let me know when you've run this through (if not already?) and I can get started on re-processing CITEseq projects.

This script has now been run so all `scpca-meta.json` files are up to date and you should be good to go to re-process CITE-seq projects.

PR #176, opened 2023-09-25
Update script for adding new fields to `scpca-meta.json` to include assay ontology and submitter cell types

### LINKED ISSUE #175: Add assay ontology to existing `scpca-meta.json` files
In https://github.com/AlexsLemonade/scpca-nf/pull/453, we are adding another term to `scpca-meta.json` files that are present in the `checkpoints` folder. Before we re-process samples to add in AnnData or cell typing, we will want to make sure the contents of `scpca-meta.json` for already processed samples match what's to be expected. 

We already have a script, [`add-refs-scpca-meta.py`](https://github.com/AlexsLemonade/alsf-scpca/blob/main/scripts/add-refs-scpca-meta.py), that reads in existing `scpca-meta.json` files and outpus a modified file with any new additions (currently just `ref_mito` and `ref_fasta_index`. We should be able to add a check for the presence of the `assay_ontology_term_id` and add it if it's not present.

allyhawkins:
Closes #175 

Since we have now added two new terms to the initial `scpca-meta.json` in `scpca-nf`, we need to update the existing checkpoints files to include those additions before we re-process for AnnData or cell type additions. 

- I updated the existing `add-refs-scpca-meta.py` to be named `add-fields-scpca-meta.py`. This script can now be used to add any new fields we may need to as we continue development of `scpca-nf`. The readme and documentation for the script usage has now been updated to reflect that. 
- I added a list of new fields that are being added. If all of them are present then no modifications are performed. Otherwise there is a check for each of the new fields individually. Then for each missing one it is added into the meta before exporting. 
- I also am checking to make sure that the submitter cell types file is up to date. I'm envisioning us populating that column as we receive new submissions. Then before we process we should be able to just run this script to update that field to match what we have in the metadata file rather than having to re-map.

I tested this with a library in `scpca/processed` and things worked as expected. Once this is approved, I'll run this for the entire `scpca-prod/checkpoints` folder.

### DIFF
--- a/scripts/add-fields-scpca-meta.py
+++ b/scripts/add-fields-scpca-meta.py
@@ -26,19 +26,19 @@ import pandas
 # Parse command line arguments
 parser = argparse.ArgumentParser()
 parser.add_argument(
-    '--library_file',
-    default='s3://ccdl-scpca-data/sample_info/scpca-library-metadata.tsv',
-    help='path or URI to library data file TSV'
+    "--library_file",
+    default="s3://ccdl-scpca-data/sample_info/scpca-library-metadata.tsv",
+    help="path or URI to library data file TSV",
 )
 parser.add_argument(
-    '--bucket',
-    default='nextflow-ccdl-results',
-    help='S3 bucket where results files are located'
+    "--bucket",
+    default="nextflow-ccdl-results",
+    help="S3 bucket where results files are located",
 )
 parser.add_argument(
-    '--checkpoints_prefix',
-    default='scpca/processed/checkpoints',
-    help='directory containing checkpoint files to modify'
+    "--checkpoints_prefix",
+    default="scpca/processed/checkpoints",
+    help="directory containing checkpoint files to modify",
 )
 args = parser.parse_args()
 
@@ -46,23 +46,24 @@ args = parser.parse_args()
 library_df = pandas.read_csv(args.library_file, sep="\t", keep_default_na=False)
 
 # remove any extra / at the end
-bucket = args.bucket.strip('/')
-checkpoints_prefix = args.checkpoints_prefix.strip('/')
+bucket = args.bucket.strip("/")
+checkpoints_prefix = args.checkpoints_prefix.strip("/")
 
 # define techs for setting up checkpionts directories for different modalities
 sc_techs = ["10Xv2", "10Xv2_5prime", "10Xv3", "10Xv3.1"]
-bulk_techs = ['single_end', 'paired_end']
-spatial_techs = ['visium']
-demux_techs = ['cellhash_10Xv2', 'cellhash_10Xv3', 'cellhash_10Xv3.1']
+bulk_techs = ["single_end", "paired_end"]
+spatial_techs = ["visium"]
+demux_techs = ["cellhash_10Xv2", "cellhash_10Xv3", "cellhash_10Xv3.1"]
 
 # go through every run id and modify scpca-meta.json file if present
 for run in library_df.itertuples():
-
     print(f"Processing {run.scpca_run_id}")
 
     # depending on the technology for that run, define the directory where scpca-meta.json is located
     if run.technology in sc_techs:
-        checkpoint_folder = f"{checkpoints_prefix}/rad/{run.scpca_library_id}/{run.scpca_run_id}-rna"
+        checkpoint_folder = (
+            f"{checkpoints_prefix}/rad/{run.scpca_library_id}/{run.scpca_run_id}-rna"
+        )
     elif run.technology in bulk_techs:
         checkpoint_folder = f"{checkpoints_prefix}/salmon/{run.scpca_library_id}"
     elif run.technology in spatial_techs:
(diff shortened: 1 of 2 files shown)


scripts/add-fields-scpca-meta.py
```diff
@@ -81,12 +87,24 @@
         print(f"No scpca-meta.json file for {run.scpca_run_id}")
         continue
 
-    # add mito file and ref fasta index if not already present
-    if all(key in results_meta for key in ('mito_file', 'ref_fasta_index')):
-        print(f"All keys are present, no updates to scpca-meta.json for {run.scpca_run_id}")
+    # create a list of new fields to check for
+    new_fields = ['mito_file', 'ref_fasta_index', 'assay_ontology_term_id', 'submitter_cell_types_file']
```
jashapiro, line 91: For easier updates later, we might make this into a dict:
```suggestion
    new_fields = {
        'mito_file': "s3://scpca-references/homo_sapiens/ensembl-104/annotation/Homo_sapiens.GRCh38.104.mitogenes.txt",
        'ref_fasta_index': "homo_sapiens/ensembl-104/fasta/Homo_sapiens.GRCh38.dna.primary_assembly.fa.fai",
        'assay_ontology_term_id', run.assay_ontology_term_id,
        'submitter_cell_types_file', run.submitter_cell_types_file
    }
```

We could then use a loop to set all of the `results_meta` values without all of the `if` statements:

```
for key, default in new_fields.items():
   results_meta.setdefault(key, default)
```

This uses `.setdefault()` which does nothing if the key exists, but updates the dictionary if it does. (Well, technically it also returns the value, but we don't need to capture that.)

This doesn't handle the case of updating the cell types file, but as that breaks the pattern, we might want to separate it more cleanly anyway.

jashapiro (commented):
I suggested a simplification, but I also wanted to comment about something I was thinking separately. 

Namely: The `assay_ontology_term_id` and `tech` columns are presumably relatable by a simple dict lookup. I think it makes sense to keep it in the file, but we ought to be able to handle those lookups in the case that the value in the table is `NA`. I'm not sure if this is the right place for it, or if it should be in the workflow itself? Or (and this is probably the correct answer) it should be in the metadata check scripts: Since we _can_ check that the two fields agree, we probably _should_.

allyhawkins:
I updated this to use the dictionary that you proposed. I also took out the submitter cell types file check and have that separate. 

> Namely: The assay_ontology_term_id and technology columns are presumably relatable by a simple dict lookup. I think it makes sense to keep both in the metadata file, but we ought to be able to handle those lookups in the case that the ontology value in the table is NA. I'm not sure if this is the right place for it, or if it should be in the workflow itself? Or (and this is probably the correct answer) it should be in the metadata check scripts: Since we can check that the two fields agree, we probably should.

I definitely like the idea of having a check in the workflow itself for NA terms for assay ontology and tech version! We could do that with the other library fields too. 
I'm not sure what you mean about having it in the metadata checks? We don't have checks between `scpca-library-metadata.json` and `scpca-meta.json` for each library?

jashapiro:
> I'm not sure what you mean about having it in the metadata checks? We don't have checks between `scpca-library-metadata.json` and `scpca-meta.json` for each library?

I meant as a step in [check_metadata.py](https://github.com/AlexsLemonade/ScPCA-admin/blob/main/sample_info/check_metadata.py)

scripts/add-fields-scpca-meta.py
```diff
@@ -81,14 +87,28 @@
...
+        print(f"All fields are present, no updates to scpca-meta.json for {run.scpca_run_id}")
     else:
-        results_meta['mito_file'] = "s3://scpca-references/homo_sapiens/ensembl-104/annotation/Homo_sapiens.GRCh38.104.mitogenes.txt"
-        results_meta['ref_fasta_index'] = "homo_sapiens/ensembl-104/fasta/Homo_sapiens.GRCh38.dna.primary_assembly.fa.fai"
+        # update missing fields
+        for key, default in new_fields.items():
+            results_meta.setdefault(key, default)
+
+        # update submitter cell types file if it's missing or if the value doesn't match the metadata file
+        if 'submitter_cell_types_file' not in results_meta.keys() or results_meta['submitter_cell_types_file'] != run.submitter_cell_types_file:
```
jashapiro, line 107: Two "pythonic" options here:

The first is a smaller change that takes advantage of the fact that `in` always checks a dictionary's keys:
```
        if 'submitter_cell_types_file' not in results_meta or results_meta['submitter_cell_types_file'] != run.submitter_cell_types_file:
```

The second is to just use the fact that `get` returns `None` if there is no corresponding key:
```suggestion
        if results_meta.get('submitter_cell_types_file') != run.submitter_cell_types_file:
```

scripts/add-fields-scpca-meta.py
```diff
@@ -81,14 +87,28 @@
...
+            results_meta['submitter_cell_types_file'] = run.submitter_cell_types_file
+
 
-    # copy updated json file
+        # copy updated json file
```
jashapiro, line 111:
```suggestion
    # copy updated json file
```

allyhawkins:
> I meant as a step in [check_metadata.py](https://github.com/AlexsLemonade/ScPCA-admin/blob/main/sample_info/check_metadata.py)

Right, but what exactly are you checking? That script checks matches between the sample and the library metadata, but the `assay_ontology_term_id` is only in the library metadata?

jashapiro (approved):
LGTM, with one simplification suggestion. I wouldn't complain if this were run through `black` as well, just for future consistency.

jashapiro:
> > I meant as a step in [check_metadata.py](https://github.com/AlexsLemonade/ScPCA-admin/blob/main/sample_info/check_metadata.py)
> 
> Right, but what exactly are you checking? That script checks matches between the sample and the library metadata, but the `assay_ontology_term_id` is only in the library metadata?

I was thinking to make sure that we had a correct match between the `technology` and `assay_ontology_term_id` as a check for data entry errors.

allyhawkins:
I ran this on all the libraries so all checkpoints files should now be up to date for future processing.

PR #177, opened 2023-11-02
Add entries for `ref_assembly` and `star_index` to `scpca-meta.json` files

allyhawkins:
In working on processing projects through `scpca-nf`, I noticed that the libraries with multiplexing were not skipping genetic demuxing, even though the results files already exist and they should be skipped. When I went to look at why, it was because the `scpca-meta.json` for multiplexed samples did not contain a `ref_assembly` version. That's what we use to check to ensure we skip demultiplexing. Here, I'm updating the script we use to add in fields to the `scpca-meta.json` files to include the `ref_assembly` field. 

When I was poking around, I also noticed that we don't store the `star_index` in `scpca-meta.json` for the multiplexed libraries, but we do for all other libraries. I am also adding that as a field to include in the json files here. 

Note that I updated one of the json files and then re-ran the workflow and things were now skipping genetic demultiplexing as they should be.

### DIFF
--- a/scripts/add-fields-scpca-meta.py
+++ b/scripts/add-fields-scpca-meta.py
@@ -8,6 +8,8 @@ This includes addition of the following missing fields:
 - 'ref_fasta_index'
 - 'assay_ontology_term_id'
 - 'submitter_cell_types_file'
+- 'ref_assembly'
+- 'star_index'
 
 For all runs present in the `--library_file`, this script will check for an existing `scpca-meta.json` file in the provided `--checkpoints_prefix` on S3.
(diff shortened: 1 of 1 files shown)


jashapiro (approved):
LGTM. Thanks for catching this! 

Just to check: are we _setting_ this in the multiplexing workflow as we should?

allyhawkins:
> Just to check: are we setting this in the multiplexing workflow as we should?

Yes, we write out the `meta` object in the `vireo` process and the `meta` object should contain both of those items.

PR #178, opened 2023-11-14
Update cell type checkpoints to include the updated reference file names

allyhawkins:
Related to https://github.com/AlexsLemonade/ScPCA-admin/issues/691

We recently updated the reference file names for both `SingleR` and `CellAssign`. Before doing this, we had previously run a few samples through both `SingleR` and `CellAssign`. To skip cell typing, we check that the reference file name stored in `library_id_cellassign/scpca-meta.json` and `library_id_singler/scpca-meta.json` match the reference file names that have been passed through the workflow via the project metadata. If we want to run the projects through again and skip running `CellAssign` for samples that already have `CellAssign` results, then these reference files need to be updated in the `scpca-meta.json` files. 

Here I'm adding a script that specifically updates the cell type `scpca-meta.json` files to make sure that the reference file names match what's in `scpca-project-celltype-metadata.tsv`. It's mostly modeled after the script we use for updating the mapping related `scpca-meta.json` files, but here we need to update two checkpoint files per library, one for each method. 

Also we want to account for values that may already be there, but with a different file name. So we directly compare what's in the checkpoint file vs. what's in the metadata file and update accordingly. Additionally, if `NA` is in the project metadata, then we don't set the path and fill with NA. Although this shouldn't really affect any of the files that will get updated here. I removed all old `SingleR` files, so in reality only the `CellAssign` results are getting updated, and they only exist if there was a reference file in the first place. 

I've tested this with runs in `scpca/processed`. Once this gets approved I'll run for `scpca-prod`.

### DIFF
--- a/scripts/README.md
+++ b/scripts/README.md
@@ -6,6 +6,10 @@ The files in this folder were utilized to create or adjust the `scpca-meta.json`
 In this script, the old results from mapping are copied over from the `internal` to `checkpoints` directory.
 Additionally, the `scpca-meta.json` file was created and saved to the `checkpoints` directory for each run to track processing information.
 
-2. `add-fields-scpca-meta.py`: This script specifically updates existing `scpca-meta.json` files within the `checkpoints` directory.
+2. `add-fields-scpca-meta.py`: This script specifically updates existing `scpca-meta.json` files within the `checkpoints` directory for mapping steps.
 For any libraries processed through `scpca-nf v0.5.1` or earlier, the `scpca-meta.json` file is modified to include additional fields that have been added in new release of `scpca-nf`.
 This includes `ref_mito`, `ref_fasta_index`, `assay_ontology_term_id`, and `submitter_cell_types_file`.
+
+3. `add-celltype-fields-scpca-meta.py`: This script specifically updates existing `scpca-meta.json` files in the `checkpoints/celltype` directory.
+For any libraries that have been processed through `scpca-nf v0.6.2` or earlier, the `scpca-meta.json` files created for cell type annotation are modified to include the most up-to-date reference file names.
+This includes ensuring that the entries for `singler_model_file` and `cellassign_reference_file` reflect the file names stored in the ScPCA project cell type metadata.
--- a/scripts/add-celltype-fields-scpca-meta.py
+++ b/scripts/add-celltype-fields-scpca-meta.py
@@ -1,0 +1,127 @@
+#!/usr/bin/env python3
+
+"""
+
+The purpose of this script is to update the the 'scpca-meta.json' checkpoint files for cell typing of already processed libraries to be up to date with changes in `scpca-nf`.
+This includes addition of the following missing fields in the celltype checkpoints:
+- `singler_ref_version`
+- `cellassign_ref_version`
+
+For all runs present in the `--library_file`, this script will check the `celltype` folder for an existing `scpca-meta.json` file in the provided `--checkpoints_prefix` on S3.
+Both the `scpca-meta.json` for SingleR and CellAssign will be updated.
+If the file is unavailable, the run will be skipped.
+If the file exists, the JSON is loaded, and the missing fields are added if they are not already present.
+To run this script for modifying the cell type `scpca-meta.json` files from runs that have already been processed for production do:
+
+python add-celltype-fields-scpca-meta.py --checkpoints_prefix "scpca-prod/checkpoints"
+
+"""
+
+import argparse
+import json
+import boto3
+import pandas
+
+# Parse command line arguments
+parser = argparse.ArgumentParser()
+parser.add_argument(
+    "--library_file",
+    default="s3://ccdl-scpca-data/sample_info/scpca-library-metadata.tsv",
+    help="path or URI to library data file TSV",
+)
+parser.add_argument(
+    "--project_celltype_file",
+    default="s3://ccdl-scpca-data/sample_info/scpca-project-celltype-metadata.tsv",
+    help="path to URI to project cell type data file TSV",
+)
+parser.add_argument(
+    "--bucket",
+    default="nextflow-ccdl-results",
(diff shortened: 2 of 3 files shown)


scripts/README.md
```diff
@@ -6,6 +6,10 @@ The files in this folder were utilized to create or adjust the `scpca-meta.json`
 In this script, the old results from mapping are copied over from the `internal` to `checkpoints` directory.
 Additionally, the `scpca-meta.json` file was created and saved to the `checkpoints` directory for each run to track processing information.
 
-2. `add-fields-scpca-meta.py`: This script specifically updates existing `scpca-meta.json` files within the `checkpoints` directory.
+2. `add-fields-scpca-meta.py`: This script specifically updates existing `scpca-meta.json` files within the `checkpoints` directory for mapping steps.
 For any libraries processed through `scpca-nf v0.5.1` or earlier, the `scpca-meta.json` file is modified to include additional fields that have been added in new release of `scpca-nf`.
 This includes `ref_mito`, `ref_fasta_index`, `assay_ontology_term_id`, and `submitter_cell_types_file`.
+
+3. `add-celltype-fields-scpca-meta.py`: This script specifically updates existing `scpca-meta.json` files in the `checkpoints/celltype` directory.
+For any libraries that have been processed through `scpca-nf v0.6.2` or earlier, the `scpca-meta.json` files created for cell type annotation are modified to include the most up-to-date reference file names.
```
jashapiro, line 14: I just want to rephrase this to confirm it will only update files names for names that were used before the current naming scheme. We don't update if the reference file was updated and an older reference was used.

scripts/add-celltype-fields-scpca-meta.py
```diff
@@ -0,0 +1,131 @@
...
+library_df = pandas.read_csv(args.library_file, sep="\t", keep_default_na=False)
+
+# Read in project celltype file
+project_df = pandas.read_csv(
+    args.project_celltype_file, sep="\t", keep_default_na=False
+)
+
+# remove any extra / at the end
+bucket = args.bucket.strip("/")
+checkpoints_prefix = args.checkpoints_prefix.strip("/")
```
jashapiro, line 59: Duplicated below (keeping the copy closer to when it is used)
```suggestion
```

scripts/add-celltype-fields-scpca-meta.py
```diff
@@ -0,0 +1,131 @@
...
+        ):
+            print(
+                f"All fields are present, no updates for {run.scpca_run_id} to {checkpoint_file}"
+            )
+        else:
+            # before modifying, make sure that the entries in the project metadata aren't NA
+            if run.singler_ref_file != "NA":
+                checkpoint_meta["singler_model_file"] = new_fields["singler_model_file"]
+            else:
+                checkpoint_meta["singler_model_file"] = "NA"
```
jashapiro, line 120: Reversing the logic here makes it a bit more clear?
```suggestion
            if run.singler_ref_file == "NA":
                checkpoint_meta["singler_model_file"] = "NA"
            else:
                checkpoint_meta["singler_model_file"] = new_fields["singler_model_file"]
```

jashapiro: Related to my first comment: I think we actually want to be more careful here, and only update if the old file had the old format. If the file name includes a version string, I think we want to _not_ update it, to save us from ourselves in the future. We should not need to run this in the future, but if we do, we don't want to accidentally change any version strings.

jashapiro (commented):
The structure here seems good, but I think we want to be a bit more strict about what files names we change, so that we don't accidentally run this in the future and update things we don't want to update. Specifically, we should check that the file name is of the "old" format before updating, so we don't accidentally update version numbers in the future when we don't really want to.

allyhawkins:
> The structure here seems good, but I think we want to be a bit more strict about what files names we change, so that we don't accidentally run this in the future and update things we don't want to update. Specifically, we should check that the file name is of the "old" format before updating, so we don't accidentally update version numbers in the future when we don't really want to.

This is a great point! I went ahead and added a check for a version string. If that's present in either of the filenames that are in `scpca-meta.json`, then no updates are made. I also updated the description of the script at the top of the file to mention that.

scripts/README.md
jashapiro, line 14:
```suggestion
For any libraries that have been processed through `scpca-nf v0.6.2` or earlier, the `scpca-meta.json` files created for cell type annotation are modified to include the current reference file names.
```

scripts/add-celltype-fields-scpca-meta.py
```diff
@@ -0,0 +1,157 @@
...
+            and checkpoint_meta["cellassign_reference_file"]
+            == new_fields["cellassign_reference_file"]
+        ):
+            print(
+                f"All fields are present, no updates for {run.scpca_run_id} to {checkpoint_file}"
+            )
+        else:
+            # find if versions are in the existing checkpoints
+            singler_version_present = re.search(
+                r"(\d{1}-\d{2}-\d{1}_model.rds$)", checkpoint_meta["singler_model_file"]
```
jashapiro, line 119: Making the regex less strict... Since version numbers may not always be 1-2-1 digits. Just take any combination of numbers and dashes.
```suggestion
                r"([-\d]+_model.rds$)", checkpoint_meta["singler_model_file"]
```

scripts/add-celltype-fields-scpca-meta.py
```diff
@@ -0,0 +1,157 @@
...
+            )
+            cellassign_version_present = re.search(
+                r"(20\d{2}-[01]\d{1}-\d{2}.tsv$)",
```
jashapiro, line 122: Don't really need the `{1}` as long as I am at it...
```suggestion
                r"(20\d{2}-[01]\d-\d{2}.tsv$)",
```

jashapiro (approved):
LGTM, with a few little changes to one of the regexes.

PR #179, opened 2023-12-18
Add back priority queue

jashapiro:
This PR adds a priority queue to our nextflow batch configs. 

Mostly this was uncommenting previous code, but I also changed the ondemand queue to use the auto-scaling ebs volumes that are part of the main queue that we use.

I had to be a bit manual with the `apply`, as I didn't have the current state file, but I put the current state file after applying my changes in S3.

Requesting @davidsmejia for review since @arkid15r is out.


davidsmejia (approved):
LGTM

PR #180, opened 2023-12-21
Add more vcpus to nextflow compute

jashapiro:
This PR is just to bump the max vcpus we allow, which will allow a bit more speed when processing large projects.


davidsmejia (approved):
LGTM

PR #182, opened 2024-01-23
Expand compute and add S3 state

jashapiro:
Here I expanded the compute available for our nextflow batch environments.

After talking with David, I realized that we can set up storing the terraform state file on S3 directly, rather than how I had been doing that quite manually, so I set that up as part of the config here as well.


arkid15r (approved):
LGTM

PR #183, opened 2024-03-01
Use named instance types for batch

jashapiro:
Here I am switching from using the "optimal" instances, which are all 4-series, to named instance types for our batch compute environments. This is related to https://github.com/AlexsLemonade/scpca-nf/issues/709 where we may need more memory than available in the R4 instances. 

I also switched to the new `SPOT_PRICE_CAPACITY_OPTIMIZED` allocation strategy where applicable, because that might control costs a bit better, especially given the fact that we are now allowing more expensive instance types.

I wasn't sure whether it was worth including both Intel and AMD instance types, but it doesn't seem like it should hurt?

I have not yet applied this change; I am waiting for review of my choices here.

### DIFF
--- a/aws/nextflow-compute.tf
+++ b/aws/nextflow-compute.tf
@@ -18,7 +18,7 @@ resource "aws_batch_compute_environment" "nf_spot" {
     instance_type = [
       "optimal",
     ]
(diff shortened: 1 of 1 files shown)


jaclyn-taroni (approved):
👍

PR #187, opened 2025-05-20
Use cellranger 9.0.1 in cellranger container

### LINKED ISSUE #186: Update cell ranger container to use v9.0.0
We will want to create a container that has `cellranger multi`. We could create a new container or just update our existing container with cellranger to use v9.0.0. That is the version that appears to [support Flex data](https://10x.vercel.app/support/software/cell-ranger/latest/release-notes/cr-release-notes). 

I don't think we actively use any workflows with the cell ranger container, but we did use it for benchmarking, so if we ever need to regenerate benchmarking data we would have to revert the container back. Because of this, maybe we should just create a different one.

allyhawkins:
Closes #186 

Here I just updated the version of `cellranger` to use 9.0.1 which is the latest version and has support for 10X flex with `cellranger multi`. To do this, I followed the instructions we had written to download the software and build the image and push to ECR. ECR should now have a new tagged version with 9.0.1 and the latest tag should also use that version.

### DIFF
--- a/images/cellranger/README.md
+++ b/images/cellranger/README.md
@@ -4,7 +4,7 @@ This folder contains a Dockerfile for the cellranger analysis.
 ## Building the image
 
 In order to build this image, the cellranger software and bcl2fastq source code components must be downloaded separately to comply with licensing, and should be placed in this folder (`images/cellranger`).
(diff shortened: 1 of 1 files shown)


jashapiro (approved):
LGTM

PR #188, opened 2025-05-28
Update base image to alma linux for cell ranger

jashapiro:
Since CentOS 7 is extremely old and unsupported, I decided to see how migrating the base image for our cellranger  image would go. I chose [Alma Linux](https://almalinux.org), which seems to be the overall successor to CentOS as far as stable RPM-based images go. 

The transition seemed to go smoothly. I updated the installation instructions just a bit to do the decompression of the Cell Ranger directory outside the Dockerfile, and that (with the base change, presumably) made the image about 2 GB smaller. 

I tested that the image works, and `cellranger` runs as expected using cellranger `testrun`, which it seems to (on the server, not on an arm mac).

If this seems reasonable to you, I can push it to AWS for further testing.

### DIFF
--- a/images/cellranger/README.md
+++ b/images/cellranger/README.md
@@ -4,7 +4,9 @@ This folder contains a Dockerfile for the cellranger analysis.
 ## Building the image
 
 In order to build this image, the cellranger software and bcl2fastq source code components must be downloaded separately to comply with licensing, and should be placed in this folder (`images/cellranger`).
-- The current version of cellranger is `cellranger-9.0.1.tar.gz` and can be downloaded from [10X Genomics Website](https://10x.vercel.app/support/software/cell-ranger/downloads#download-links) after agreeing to their license terms.
+- The current version of cellranger is `cellranger-9.0.1.tar.xz` and can be downloaded from [10X Genomics Website](https://10x.vercel.app/support/software/cell-ranger/downloads#download-links) after agreeing to their license terms.
+  - Download the `cellranger-9.0.1.tar.xz` file and extract it using the `tar -xJf cellranger-9.0.1.tar.xz` command.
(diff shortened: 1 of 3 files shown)


images/cellranger/README.md
```diff
@@ -4,7 +4,9 @@ This folder contains a Dockerfile for the cellranger analysis.
 ## Building the image
 
 In order to build this image, the cellranger software and bcl2fastq source code components must be downloaded separately to comply with licensing, and should be placed in this folder (`images/cellranger`).
-- The current version of cellranger is `cellranger-9.0.1.tar.gz` and can be downloaded from [10X Genomics Website](https://10x.vercel.app/support/software/cell-ranger/downloads#download-links) after agreeing to their license terms.
+- The current version of cellranger is `cellranger-9.0.1.tar.xz` and can be downloaded from [10X Genomics Website](https://10x.vercel.app/support/software/cell-ranger/downloads#download-links) after agreeing to their license terms.
+  - Download the `cellranger-9.0.1.tar.xz` file and extract it using the `tar -xJf cellranger-9.0.1.tar.xz` command.
```
allyhawkins, line 8:
```suggestion
  - Download the `cellranger-9.0.1.tar.xz` file and extract it using the `tar -xzf cellranger-9.0.1.tar.xz` command.
```
Is this what you meant? I didn't think J was an option?

allyhawkins (approved):
Thanks for doing this! Seems reasonable to me.

images/cellranger/README.md
jashapiro: `J` is to for `.xz` archives. But really it could just be `-xf` and tar should autodetect, so that's what I'll do. 

```suggestion
  - Download the `cellranger-9.0.1.tar.xz` file and extract it using the `tar -xf cellranger-9.0.1.tar.xz` command.
```

PR #191, opened 2025-09-10
Add infercnv_gene_order to add fields script

### LINKED ISSUE #190: Add infercnv_gene_order file to checkpoint meta
While working on https://github.com/AlexsLemonade/scpca-nf/issues/965, got errors that the `infercnv_gene_order` file is `null` which is because, indeed, this field is not in the checkpoint meta that gets read in to resumed runs. 

We'll need to update this script to include adding the field `infercnv_gene_order`:
https://github.com/AlexsLemonade/alsf-scpca/blob/main/scripts/add-fields-scpca-meta.py

---

Looking at this script for the first time in a minute too, I'll also raise the point that we might want (need?) to spruce up this script more generally to allow for different reference/ensembl versions; this would be a separate issue though!
https://github.com/AlexsLemonade/alsf-scpca/blob/9428f29f32f97b0da6bae71d6a9d8932a1615a23/scripts/add-fields-scpca-meta.py#L116-L122

sjspielman:
Closes #190
Does what it says!

As part of review, note also I included a blurb in the issue here that maybe we need more general updates to the script? Let me know if I was barking up the right tree there and we'll indeed open another issue to circle back to.

### DIFF
--- a/scripts/add-fields-scpca-meta.py
+++ b/scripts/add-fields-scpca-meta.py
@@ -10,6 +10,7 @@ This includes addition of the following missing fields in the mapping checkpoint
 - 'submitter_cell_types_file'
 - 'ref_assembly'
 - 'star_index'
+- `infercnv_gene_order`
 
 For all runs present in the `--library_file`, this script will check for an existing `scpca-meta.json` file in the provided `--checkpoints_prefix` on S3.
 If the file is unavailable, the run will be skipped.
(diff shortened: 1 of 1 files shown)


allyhawkins (approved):
LGTM, you'll need to run it for the checkpoints in `scpca/processed` folder and then before we process everything, you'll need to run it in the `scpca-staging` folder. You can do that now, or file an issue to make sure we do that before we process samples.

sjspielman:
I'll file an issue for it. For now, I manually updated the checkpoints in `scpca/processed` for the sample I'm using for testing (you guessed it, sample 1).
Worth noting that my draft infercnv module is at least running - the infercnv process starting running (not submitted, _running_) about 67 minutes ago and it's still chugging along...so good thing we're using checkpoints!

PR #192, opened 2025-09-10
Document that last PRs change

sjspielman:
Quick followup - adding those columns (and other missing ones) to the README!

Also, question for reviewer - what version of `s3fs` do you have? When running this script, I am getting messages that this dependency is needed, but the version that gets installed from conda (0.4.2) gives this concerning warning:
```
UserWarning: Your installed version of s3fs is very old and known to cause
severe performance issues, see also https://github.com/dask/dask/issues/10276

To fix, you should specify a lower version bound on s3fs, or
update the current installation.
```

I don't see an environment file in this repo for pinning anything, so I figure I'll ask what you've had success with and go from there?


allyhawkins (approved):
👍

sjspielman:
Wow, what on earth is conda doing?! pip it is!

PR #196, opened 2026-07-06
Remove legacy AWS infrastructure

jashapiro:
Closes https://github.com/AlexsLemonade/ScPCA-admin/issues/1446

Rather than fully deleting these files, I left them with all of the resources commented out, just in case we want to refer back to how we were doing things like the auto-expanding drives. While we could of course extract them from git history, that seems like it would be harder.

I have not yet fully applied this, though I did delete some elements already... I will apply it before merging.

I already removed the relevant AMI, which was probably just a bit premature.


jaclyn-taroni (approved):
👍🏻

jashapiro:
Successfully applied: all resources removed.

